PR #12, opened 2016-10-13
Single process with no children

giorgiosironi:
Executing Python from a shell script creates two processes:
```
17553 ?        Ss     0:00 /bin/bash ./bot-lax-listener.sh end2end 1
17598 ?        S      0:00  \_ python src/adaptor.py --type sqs
```
Using `exec` replaces the shell with the new process:
```
17598 ?        S      0:00  python src/adaptor.py --type sqs
```
so that when Upstart restarts it, this single process is terminated and
created again rather than leaving around orphans.

### DIFF
--- a/bot-lax-listener.sh
+++ b/bot-lax-listener.sh
@@ -8,4 +8,4 @@ set -e
 # the 'ci' in 'lax--ci'
 instance_id=$1
 
-python src/adaptor.py --type sqs --instance "$instance_id"
+exec python src/adaptor.py --type sqs --instance "$instance_id"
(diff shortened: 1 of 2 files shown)


giorgiosironi:
The install fails because of a Jenkins change, `/var/lib/jenkins-libraries-runner/workspace/aries_bot-lax-adaptor_PR-12-QAM6W2XLBGLG5PGROBJBH3K3STLN5BHBZJHOONLCCDQ65OVCQBKQ/venv/bin/python` seems to go over the kernel limits for a `#!` line at the start of `pip`.

giorgiosironi:
https://issues.jenkins-ci.org/browse/JENKINS-34564

giorgiosironi:
Ok, now it fails but for the correct reason, incomplete coverage.

PR #15, opened 2016-10-17
Feature validate to merge

gnott:
* Moved some rewriting over to validate.py (no more body_rewrite())
* More rewriting and placeholders based on the validate-demo branch
* Refactor clean() to also be applied to snippets
* research-organisms to researchOrganisms
* Placeholder data for references, which are still under development and not merged into elifetools yet


lsh-0:
well done on this, I'm grateful.

I'm waiting on some feedback from Chris, but I might be dropping the snippet handling and generating it from Lax rather than manage the double handling we see appearing here.

PR #16, opened 2016-10-19
Fixing graceful shutdown

giorgiosironi:
EDIT: it turns out I already had a pull request fixing this and other problems of restarts around.

From
https://ci--alfred.elifesciences.org/job/test-elife-lax/110/consoleFull

```
      ID: bot-lax-adaptors-start
Function: cmd.run
    Name: start bot-lax-adaptors
  Result: True
 Comment: Command "start

 Started: 11:52:22.340042
Duration: 30129.05 ms
 Changes:
          ----------
          pid:
              3412
          retcode:
              0
          stderr:
          stdout:
              bot-lax-adaptors stop/waiting
```

The graceful shutdown times out after 30 seconds and the process gets
killed with `SIGKILL`.

The reason is if there are no messages (which is 99% of the time), the
inner while loop returns to polling without yielding a new value,
never exiting.

Therefore, passing the flag to the IncomingQueue allows it to check
whether it should stop, before starting any new polling. If polling
returned something, that is always processed before attempting a new
iteration.

### DIFF
--- a/src/adaptor.py
+++ b/src/adaptor.py
@@ -244,10 +244,10 @@ except ImportError:
 #
 #
 
-def read_from_sqs(stackname='temp'):
+def read_from_sqs(instance_id='temp', **kwargs):
     "reads messages from an SQS queue, writes responses to another SQS queue"
-    incoming = sqs_adaptor.IncomingQueue('bot-lax-%s-inc' % stackname)
-    outgoing = sqs_adaptor.OutgoingQueue('bot-lax-%s-out' % stackname)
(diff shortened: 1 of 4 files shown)


giorgiosironi:
The test seems broken on `develop` too.

giorgiosironi:
I'll rebase on the current `develop` which is green.

giorgiosironi:
Now green.

lsh-0:
feel free to merge when you're ready

PR #20, opened 2016-10-20
Feature parallel processing of json generation and validation

lsh-0:
added support for running the article json generation and validation over all available cores.

the scripts `./generate-article-json.sh` and `./validate-json.sh` should continue to work as before, but faster.

### DIFF
--- a/src/tests/test_main.py
+++ b/src/tests/test_main.py
@@ -1,10 +1,9 @@
-import sys, StringIO
 import json
 from os.path import join
 from .base import BaseCase
 import main
 
-class TestArticleScrape(BaseCase):
+class ArticleScrape(BaseCase):
     def setUp(self):
         self.doc = join(self.fixtures_dir, 'elife-09560-v1.xml')
         self.doc_json = join(self.fixtures_dir, 'elife-09560-v1.json')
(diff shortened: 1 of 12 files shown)


gnott:
That sure gives my CPU a workout!

I cannot use os.symlink unfortunately in validate, if you would see your way to changing those to a file copy procedure I would be able to use it.

In generate also to import conf.py for the folder names so they're consistent?

Otherwise this will be a good time saver when wanting to generate and test the entire set of articles.

src/tests/test_main.py
```diff
@@ -35,16 +34,10 @@ def test_render_single(self):
...
-        strbuffer = StringIO.StringIO()
-        sys.stdout = strbuffer
-        try:
-            main.main(self.doc) # writes article json to stdout
-            results = json.loads(strbuffer.getvalue())
-            self.assertTrue('article' in results)
-            self.assertTrue('journal' in results)
-        finally:
-            sys.stdout = _orig
+        results = main.main(self.doc) # writes article json to stdout
```
giorgiosironi: ah, the wins of designing for testability

giorgiosironi:
Would like to run this locally to see it in action, but starting `./generate-article-json.sh` I get:
```
Traceback (most recent call last):
  File "src/generate_article_json.py", line 7, in <module>
    import main as scraper
  File "/home/giorgio/code/bot-lax-adaptor/src/main.py", line 254, in <module>
    ('titlePrefix', [jats('title_prefix')]),
  File "/home/giorgio/code/bot-lax-adaptor/src/main.py", line 141, in jats
    actual_func = getattr(parseJATS, funcname)
AttributeError: 'module' object has no attribute 'title_prefix'
```
which is yesterday's error. How can we fix this to run this build? Also the Alfred build has a different failure.

lsh-0:
@giorgiosironi , I think I fixed that by nuking my venv and re-installing. The schema repo should be automatically updated when you call the validate script so no problems there unless it's on a different branch.

@gnott yes, I forgot to fix up those conf settings in generate. I'll clean that up now. I'll also do an os detection for windows and get files copied instead of symlinked - this will be Disk IO  intensive and probably slow the generation down some unless you're on an ssd

lsh-0:
@gnott support for copying files instead of symlinking if windows detected. let me know if this works for you

gnott:
Much better here, thanks. It's good to see validate() return something too when testing. :+1: 

I will wait for this to be merged before I create a PR with references data, because I may have some conflicts to resolve first.

PR #22, opened 2016-10-26
Feature references part 1

gnott:
Basic functionality to use the parser references_json() function. It also has validation rewrites and placeholders to get the most valid XML possible at this time.

No need to merge, but it can be if you like it @lsh-0.

Still to do with references is
- adding authors and all other people
- converting references to "unknown" type if they will not pass validation. That should make it so placeholders are no longer required, so this is an interim step to that goal

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -288,7 +288,7 @@ VOR.update(OrderedDict([
     ('relatedArticles', [jats('related_article'), related_article_to_related_articles]),
     ('digest', [jats('digest_json')]),
     ('body', [jats('body')]), # ha! so easy ...
-    ('references', [jats('references')]),
+    ('references', [jats('references_json')]),
     ('decisionLetter', [jats('decision_letter')]),
     ('authorResponse', [jats('author_response')]),
 ]))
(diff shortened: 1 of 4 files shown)


lsh-0:
I'll leave the merging of changes to the scraper up to you, but let me know if you ever want me to cast my eye over anything. I think we're all on the same page now regarding where logic should live.

gnott:
Thanks Luke. I think I will merge so this code doesn't go stale. The main difference is now using references_json() from the parser in main.py  I just didn't want it to cause any havoc when merging this part 1, although I have added rewrites and placeholders to the best of my ability to account for data that is not passing validation naturally. It will take some time to produce reference data without placeholder assistance.

lsh-0:
>  I just didn't want it to cause any havoc when merging this part 1,

if your code is passing tests it should be safe to merge into develop. Alfred will run develop through integration tests then merge into master if everything passes.

if havoc does occur we haven't written enough tests or our infrastructure isn't up to snuff.

Don't be afraid of havoc.

PR #30, opened 2016-11-08
added pdf url handling

lsh-0:


gnott:
Quick comment, PDF files are versioned and the file names are reflected there. The cdn.elifesciences.org version of PDF files is old and my planned (delayed) coding updates will actually stop updating that bucket. You might not be done or maybe this is a temporary configuration? In that case you can ignore my comments!

lsh-0:
that is why there are these pull requests, buddy :)

so, api.elifesciences.org is returning the pdf from the cdn - I'm just using it's logic. Is there a different place I should be looking to serve pdf files from other than the cdn? I don't recall seeing any versioned files in there yesterday when I took a quick look.

I'd be grateful if you could give me the requirements to finish this one off.

lsh-0:
ok, we've figured out what you mean. a ticket has been opened to deal with multiple CDNs. that blocks this PR.

lsh-0:
I've confirmed all articles except corrections have a pdf. merging.

PR #34, opened 2016-11-15
No more author rewriting with phone number format moved to elifetools…

gnott:
… parser. Also do not count __main__ functions in coverage report to stay above 80% coverage.


giorgiosironi:
No worries, `main` is hit hard by all the end2end tests, which do no track coverage.

PR #37, opened 2016-11-21
Feat backfill lax

lsh-0:
a script has been created to back-fill lax.

it uses the elife-article-xml repository, generates the article-json, validates it and sends the valid json to lax as an INGEST --force command.

this branch also pins the version of elifetools being used so we can get into the habit of reproducible builds. this makes everybody's lives easier, but especially @giorgiosironi 's


README.md
```diff
@@ -2,60 +2,83 @@
...
+
+### receiving messages from an AWS SQS queue
+
+This is quite eLife-specific but can be modified easily if you're a developer:
+
+    $ ./bot-lax-listener.sh
+
+## testing
+
+    $ ./test.sh
```
giorgiosironi: :+1: nice documentation

PR #43, opened 2016-11-24
Script for updating elifetools to the latest version from its develop branch

giorgiosironi:
Only execute when this is needed, normally elifetools will be pinned to the commit inside `requirements.txt`

Normally `pip freeze` should do this job, but it doesn't work with commits, only with releases.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -8,7 +8,9 @@ fi
 
 source venv/bin/activate
 
-pip uninstall -y elifetools
+if pip list | grep elifetools; then
+    pip uninstall -y elifetools
+fi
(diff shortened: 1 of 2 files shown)


giorgiosironi:
@lsh-0 I just had to manually recreate the virtualenvs of bot-lax-adaptor in prod because it was updated again, once this is green we need it to correctly update `elifetools`. I'll also try to monitor the Upstart processes so that we know if it goes down.

giorgiosironi:
Merging this to try and avoid more breakages, locally works but will try it out on end2end.

PR #44, opened 2016-11-24
updated xml and raml versions

lsh-0:


lsh-0:
ping @giorgiosironi 

once these pass tests we should update the sha of elife-tools in the `requirements.txt` file

giorgiosironi:
About elife-tools: check out https://github.com/elifesciences/bot-lax-adaptor/pull/43

giorgiosironi:
With this change:
```
14:46:15 [4-BASNE5BWTFRD5Q5CGCLU2TGU6BPE63PNPDFXVCQBHBVAGXEADCRA] Running shell script
14:46:16 + ./validate-statistics.sh validation.log
14:46:16 Input: 4908 JSON articles
14:46:16 Output valid: 4303 JSON articles
14:46:16 Output invalid: 605 JSON articles
14:46:16 Validated green,red,total: 4303,605,4908
14:46:16 No more than 91 articles should be failing validation (with placeholders)
14:46:16 INFO - 2016-11-24 14:42:25,511 - elife-23066-v1.xml.json => failed
14:46:16 INFO - 2016-11-24 14:42:25,516 - elife-22300-v1.xml.json => failed
...
```
I am inclined to merge it by setting the threshold to 605, because this is a stricter requirement and not a regression of the code. But we should be careful when doing so as this may hide other errors in those 605.

gnott:
I will take a look at the validation number. I actually didn't check a bulk validation on datasets yet (but I didn't approve that version of elifetools yet). It could also be the section title length issue if that is in effect now.

lsh-0:
> but I didn't approve that version of elifetools yet

check for errors, if it's just strictness then bump the elife-tools sha

lsh-0:
sorry. clicked wrong button.

if after checking the errors and it's just strictness checking, we could try bumping the sha for elife-tools

giorgiosironi:
Yes, if this was unclear, 605 invalid articles were due to the stricter schema now in place requiring section titles not to be empty. After Graham's fixes yesterday now we have 87 invalid so I'll even lower the threshold and then merge.

PR #47, opened 2016-11-25
Updated xml repository

giorgiosironi:


giorgiosironi:
4 articles more (valid). Second part of #44

PR #51, opened 2016-11-28
Checking out the latest api-raml

giorgiosironi:
Seeing what happens with stricter dates


giorgiosironi:
```
11:35:31 Input: 4912 JSON articles
11:35:31 Output valid: 0 JSON articles
11:35:31 Output invalid: 4912 JSON articles
11:35:31 Validated green,red,total: 0,4912,4912
```
means that the values bot-lax-adaptor is using to fill in the article JSON are not valid according to the new stricter api-raml with `Z` as timezone. Where are that specified?

lsh-0:
hm - could this be a python jsonschema problem?

lsh-0:
huh - if I read this correctly, jsonschema is checking the format of things? fml.
http://python-jsonschema.readthedocs.io/en/latest/validate/#validating-formats

PR #78, opened 2016-12-13
added a simple web interface 

lsh-0:
allows individual articles to be uploaded/generated/validated/deleted


giorgiosironi:
which port and/or virtual host it will use? (guessing on the lax--prod instance where the code is already deployed)

lsh-0:
not sure - haven't thought that far ahead yet. we'll need to update the formula, possibly password protect things.

lsh-0:
want to add a test or two before I merge this

lsh-0:
added some tests for the flask server, just enough to get us over the 80% coverage mark.

I'm not sure how much time to spend on coverage I want on this until we get some feedback from production as to whether it's useful

PR #79, opened 2016-12-14
removed the beginnings of a very bad idea

lsh-0:


lsh-0:
ping @giorgiosironi , @gnott

giorgiosironi:
For the passionate readers:
http://misko.hevery.com/code-reviewers-guide/flaw-brittle-global-state-singletons/

lsh-0:
and arguing it from an OO pov is also a bit ironic. thanks for the read though.

PR #81, opened 2016-12-17
Update elifetools and hope for less valid articles.

gnott:


giorgiosironi:
:+1: going down towards 0

PR #83, opened 2016-12-19
Remove placeholders for box content and section id, should be …

gnott:
…generally fixed by parser and silent corrections (so use latest XML repo).


gnott:
No additional invalid articles after the parser changes and including the most recent eLife XML (which includes some silently corrected articles). i.e., placeholders are removed with no consequence. I will merge this.

giorgiosironi:
Nice work @gnott!

PR #91, opened 2017-01-04
added test for expansion of relative urls within inline html

lsh-0:
added breaking test that will pass when the expansion of relative urls is fixed

pinging @giorgiosironi

### DIFF
--- a/src/tests/test_main.py
+++ b/src/tests/test_main.py
@@ -227,3 +227,22 @@ class KitchenSink(BaseCase):
             }]
         }
         self.assertEqual(tod(expected_media), tod(media))
+
+    def test_expanded_inline_urls(self):
+        result = main.render_single(self.doc, version=1)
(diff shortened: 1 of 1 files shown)


thewilkybarkid:
I'm assuming this is related to the current end2end failures (eg https://ci--alfred.elifesciences.org/job/test-journal/279/)?

lsh-0:
yah. there has been some regression and now we have relative urls again in inline html

giorgiosironi:
Yes, but we have rolled back Lax so test-journal should become green again.

giorgiosironi:
Good test by the way :+1:

giorgiosironi:
A bit outdated probably, should be rebased to be tested again.

giorgiosironi:
Looks very outdated and conflicts with `develop`; I'll close, reopen if needed.

PR #92, opened 2017-01-04
better volume calculation

lsh-0:


lsh-0:
> Along the lines I was drafting before I backtracked to create this ticket first. 
> I think jats("pub_date") returns a date_struct, so pub_date[0] is the year, but I could 
> be wrong. Just throwing that out there if it helps.

I'll take a second look. Wasn't too happy at the time with the double handling of strings for testing and time structs with parseJats

PR #103, opened 2017-01-25
Test latest elifetools and article XML.

gnott:


giorgiosironi:
From what I know, Luke has an explicit test using a kitchen sink article here: don't know how recent that is, it may be an old copy that needs updating.

gnott:
Thanks for the tip @giorgiosironi. This latest version of the elifetools parser normalises DOI values to be the same when parsing the new kitchen sink and old XML, and I updated the test scenario to get it to pass.

Updating the kitchen sink XML in this project should probably be done at some point too.

PR #121, opened 2017-02-20
Feature: pass related articles

lsh-0:


src/main.py
```diff
@@ -144,18 +144,11 @@ def display_channel_to_article_type(display_channel_list):
...
-            if doi:
-                related_articles.append(doi.split('.')[-1])
-    if len(related_articles) <= 0:
-        return None
-    return related_articles
+    # ll: [{'xlink_href': u'10.7554/eLife.09561', 'related_article_type': u'article-reference', 'ext_link_type': u'doi'}]
+    def et(struct):
+        return struct.get('xlink_href', '').rsplit('.', 1)[-1] or None
+    # ll: ['09561'] or None
+    return filter(None, map(et, related_article_list)) or None
```
giorgiosironi: why not `[]` rather than `None`?

src/main.py
```diff
@@ -453,7 +446,7 @@ def postprocess(data):
 VOR = copy.deepcopy(VOR_SNIPPET)
 VOR.update(OrderedDict([
```
giorgiosironi: so `POA` never have related articles?

giorgiosironi: (answering myself): they never have

src/main.py
lsh-0: preserving previous behaviour so tests continue to pass. I probably will change it to `[]` once I get stuck into it.

src/main.py
lsh-0: Nope ;) I updated the jira ticket with some of the facts that came out of the email conversation, one of which is that POAs don't have related articles.

lsh-0:
This is ready to merge.

PR #126, opened 2017-02-28
Feat: backfill unpublished invalid

lsh-0:
pinging @giorgiosironi , @gnott 

https://elifesciences.atlassian.net/browse/ELPP-2337

so, this is a bit of a mess, but here goes:
* this change will make the bot-lax-adaptor start sending the location of the xml to lax
* lax has a PR that will output invalid and unpublished+invalid articles and their download location if it has one
* the `backfill-unpublished.*` scripts will call this command on lax to get these problem articles and then generate requests to re-send them.

it's just a first attempt and ... I've just thought of a huge problem: backfill bypasses all of this and won't send the location info across.

bugger.

I could add some cleverness to the `validate.py` script to add the location information there, but then it will need to inspect S3 on each scrape.


src/backfill_unpublished.py
```diff
@@ -0,0 +1,28 @@
...
+    cmd = [
+        adaptor.find_lax(), # /srv/lax/manage.sh
+        "status",
+        "article-versions.invalid-unpublished.list"
+    ]
+    lax_stdout = None
+    rc, lax_stdout = utils.run_script(cmd)
+
+    requests = []
+    for av in json.loads(lax_stdout)['article-versions.invalid-unpublished.list']:
```
giorgiosironi: the `article-versions.invalid-unpublished.list` field access looks a bit strange, what other fields are there?

giorgiosironi:
I think I understand that the `backfill.sh` script, for performance reasons, doesn't go through the code in `adaptor.py` essentially.

Still, to be consistent it needs to populate the `location` field inside the article metadata, since it overwrites the article version and would make the stored article version lose this field (my hypothesis).

I see a couple options out of this:
1. in Lax, keep the metadata if not explicitly overwritten, copying from the previous article version when it is being overwritten. Not particularly keen on this it's complex.
2. provide a local absolute path as `location` for in `main.main` and/or `main.render_single`, maybe an additional argument to it?

In option 2, which seems more promising, `backfill.sh` will eventually call `render_single(doc, ..., location='/opt/bot-lax-adaptor/article-xml/articles/elife-12345-v1.xml')`. `adaptor` will call `render_single(doc, ..., location='https://s3.amazonaws.com/...')`. Given a version of lax, the bot lax adaptor is pinned and has a pinned elife-article-xml reference that should always contain that XML file, so absolute local paths are stable.

A variation in option 2 is to pass in the github URL, e.g. `https://raw.githubusercontent.com/elifesciences/elife-article-xml/84be49451c5227fbcb6b3c68b9a9c24e0857212d/articles/elife-25532-v2.xml`. There's even a sha in there for reproducibility, so maybe it's cleaner.

Sample life of an article:
```
- v1 ingested: location=s3, run A
- v1 ingested again: location=s3, run B
- v1 backfilled: location=s3, run B (taken from Lax and reused)
- v1 published
- v1 backfilled: location=local or github
- v2 ingested: location=s3, run C... (everything is the same for new versions)
```

src/adaptor.py
```diff
@@ -215,6 +215,11 @@ def handler(json_request, outgoing):
             if conf.SEND_LAX_PATCHED_AJSON: # makes in-place changes to the data
                 validate.add_placeholders_for_validation(article_data)
                 LOG.info("placeholders attached")
+
```
giorgiosironi: I think this should be pushed into `render_single` as the lowest common denominator between backfills and adaptor ingests; that would make explicit how to always provide a location, see my long comment.

src/backfill_unpublished.py
lsh-0: I think this is what you're after: https://github.com/elifesciences/lax/pull/99/files/2f41354d08dc436910c4c94bf59973781e78da77#diff-88497525781680406c91290667cfbce3R51

that `article-versions.invalid-unpublished.list` is just a path into a dict, in this case `{'article-versions': {'invalid-unpublished': 'list': [...]}}` that is dealt with here: https://github.com/elifesciences/lax/pull/99/files/2f41354d08dc436910c4c94bf59973781e78da77#diff-4af08ff4de97e518a6cc95efb2fe826cR28

you could do the same with `jq`, but I don't like `jq` except for really basic things.

lsh-0:
yes - I like your variation on option 2. when we backfill, the s3 location will be replaced with the github location but until then, it will have it's original s3 location. And I think I can push that snippet into `render_single` without too much fuss.

we'll still need to deal with the current crop of invalid+unpublished articles, but there are very few of them.

gnott:
Sorry it seems we commit around the same hours of the day now, and I just merged in using new elifetools version, hence the conflict.

lsh-0:
@gnott , all good, already resolved

lsh-0:
ok - this is doing the long test now. I won't be around for testing to end, so if it completes, and you guys are happy with the code, please merge and do a backfill and get this one off our backs: https://elifesciences.atlassian.net/browse/ELPP-2337

it *won't* handle unpublished+invalid articles just yet, but this adds support for finishing the backfill-unpublished.sh script off in another PR

lsh-0:
With the refactor I've just done it may be possible to get rid of the thread-local hack now. I'll take a look at this today.

lsh-0:
after yanking `threading.local`, we have two new problems:
* cdn link: https://github.com/elifesciences/bot-lax-adaptor/pull/126/files#diff-0c81b9c94c68ea34c497c1fa52daee56R139
* instance id: https://github.com/elifesciences/bot-lax-adaptor/pull/126/files#diff-da4832ab08d706baa58694b0de8a2fd2R320

both could be fixed by introducing by introducing a .cfg file and having .conf tackle that.

giorgiosironi:
Nice!
Yes, a `.cfg` file would hold information for both environment name and for the CDN to link to. This will make the code more agnostic with respect to eLife's idiosyncrasies, good sign.

src/main.py
```diff
@@ -481,12 +457,49 @@ def mkdescription(poa=True):
...
+        return path
+
+    # who knows what this path is ...
+    LOG.warn("scraping article content in a non-repeatable way. please don't send the results to lax")
+    return path
+
+def render_single(doc, **ctx):
     try:
-        setvar(**overrides)
+        ctx['location'] = expand_location(doc)
```
lsh-0: FIXME: this nukes any explicit location value that is passed in.

lsh-0: `ctx['location'] = ctx.get('location', expand_location(doc))` perhaps

src/main.py
```diff
@@ -170,9 +135,10 @@ def et(struct):
     return map(et, mixed_citation_list)
 
 def cdnlink(msid, filename):
-    cdn = conf.cdn(getvar('env', None)(None))
+    #cdn = conf.cdn(getvar('env', None)(None))
+    cdn = conf.cdn(env=None)
```
lsh-0: FIXME: read from config file. `return conf.cdn % kwargs`

src/adaptor.py
```diff
@@ -322,7 +317,8 @@ def bootstrap():
 
     args = parser.parse_args()
 
-    main.setvar(env=args.instance_id)
+    # TODO: how to pass this to render?
+    # main.setvar(env=args.instance_id)
```
lsh-0: `instance_id` is used for the cdn link generation. cdn link should be read from a config file and this `instance_id` argment should be removed

lsh-0:
related commit to the formula to add a `app.cfg` file with the proper environment:
https://github.com/elifesciences/lax-formula/commit/f58ece2a7366e15a5dbb45b5cfb49442fc09be74

lsh-0:
alright - I'm going to merge this in and work on the `backfill-unpublished.sh` script in a different branch.

PR #130, opened 2017-03-06
removed incorrect volume calculation.

lsh-0:
(for unpublished POA articles)

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -167,7 +167,7 @@ def category_codes(cat_list):
 def handle_isbn(val):
     return mask(to_isbn13(str(val)))
 
-def to_volume_correct(pair):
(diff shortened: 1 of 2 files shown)


gnott:
The comment that says eLife journal inception is 2011 had me confused when doing the math in my head, but I see in conf.py you have it as 2012 (which seems right).

lsh-0:
yeah, so inception = 2012 but to do the calculation and not get a volume of zero, you subtract `1` from it. the tests in `test_main.py` make this really clear

lsh-0:
absolutely. is this the end2end I should be looking at: 
https://ci--alfred.elifesciences.org/job/test-lax/

PR #132, opened 2017-03-07
Feat backfill unpublished2

lsh-0:
this is the logic to support backfills of articles in lax.

they need not be invalid or unpublished or anything like that, but there is a `backfill-unpublished.sh` script that asks lax for those articles that are unpublished (and invalid) and calls the backfill script with them.

backfill script supports two types of input: paths and simple json objects with msid, version and location attributes (that lax happily emits as part of it's `status` report).

I've also added an example backfill script to demonstrate how it should be called (`backfill-example.sh`).

The `backfill.py` script has a `--dry-run` parameter that will simply display the request it would have otherwise sent to lax. The request contains what type of action to be taken (forced ingest) the full location of the article (s3 or fs path) etc.


lsh-0:
hrm - I don't think `--ci` has permissions to talk to s3 ... ? I should probably mock that call anyway

giorgiosironi:
Yes, should be mocked, the bucket is property of the bot and doesn't exist; even if it were to exist, only the bot should talk to it in CI; mocking the HTTP call is the best choice.

PR #133, opened 2017-03-07
Do not install on startup

giorgiosironi:
Similar to https://github.com/elifesciences/lax/pull/111

### DIFF
--- a/.bot-lax-listener.sh
+++ b/.bot-lax-listener.sh
@@ -1,9 +1,4 @@
 #!/bin/bash
 set -e
 
-source venv/bin/activate
-
-# the 'ci' in 'lax--ci'
-instance_id=$1
-
 exec python src/adaptor.py --type sqs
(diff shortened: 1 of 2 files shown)


lsh-0:
yeah, and just like lax, it's assumed these scripts that aren't hidden will ensure the environment is brought into a valid state when run. there are assumptions throughout these projects that rely on this being so.

scripts like:
* `manage.sh`
* `test.sh`
* `project_tests.sh`
* `scrape-article.sh`

etc etc etc all source `install.sh`. scripts that are hidden, like:

* `.lint.sh`
* `.scrub.sh`
* `.test.sh`
* `.cc-check.sh`

make no such assumptions.

I've added a PR to lax that adds a `--skip-install` flag to `manage.sh` because it will help with the `backfill.py` work, but as this PR is I don't want it merged.

If you need the `--skip-install` functionality in bot-lax, feel free to reopen and copy+paste from there.

PR #135, opened 2017-03-08
Avoid dependency on Bitbucket when ingesting or publishing articles

giorgiosironi:
Using the new `--skip-install` flag should remove the reinstall of all
Python dependencies on every article publishing and hence this failure
mode, plus saving a few seconds in every test.


lsh-0:
I've asked the author of the rfc3339 library to publish a new version on pypi twice but he hasn't got back to me. for the time being it's just pointing to the bitbucket repo but I'll asking him one more time and then perhaps fork+publish it under a different name

PR #145, opened 2017-03-20
Use IIIF server for image or inline images, the CDN for other files.

gnott:


src/conf.py
```diff
@@ -116,6 +116,8 @@ def json_load(path):
 }
 CDN = 'https://' + CDNS_BY_ENV.get(ENV, DEFAULT_CDN)
 
+IIIF = 'https://prod--iiif.elifesciences.org/lax:%(padded-msid)s/%(fname)s'
```
giorgiosironi: similar to the `CDNS_BY_ENV` configuration in this file, this should be:
```
'https://'+ENV+'--iiif.elifesciences.org/...'`
```

src/tests/test_main.py
```diff
@@ -187,6 +189,18 @@ def test_expand_uris(self):
         ]
         self.assertEqual(main.expand_uris(msid, given), expected)
 
+    def test_expand_inline_graphics(self):
```
giorgiosironi: I'm not sure the inline graphics should go through IIIF rather than just the CDN. The  tag doesn't look like it can support a full-fledged triple IIIF viewer? The URL in `src` would redirect to the `info.json` image descriptor rather than returning the image content.

giorgiosironi: /cc @thewilkybarkid

thewilkybarkid: Yeah, they shouldn't use IIIF.

src/conf.py
gnott: Great point - I was mostly just checking for validation on this first run, and this detail will be added shortly.

src/tests/test_main.py
gnott: If the inline-graphics load from the CDN rather than the IIIF server, then that is much simpler. Is that the same for video stills, or are those still coming through IIIIF?

thewilkybarkid: Video stills are through IIIF.

thewilkybarkid: Only inline graphics from the CDN (as we can't do anything with them, and they're tiny).

src/tests/test_main.py
```diff
@@ -256,7 +269,7 @@ def tod(d):
                 "uri": main.cdnlink("00666", "elife-00666-video4-data1.xlsx"),
                 "filename": "elife-00666-video4-data1.xlsx"
             }],
-            "image": "https://static-movie-usa.glencoesoftware.com/jpg/10.7554/657/f42a609b0e61fc41798dcba3cc0c87598bd2cf9f/elife-00666-video1.jpg",
+            "image": main.iiiflink("00666", "elife-00666-video1.jpg"),
```
giorgiosironi: is journal capable of seeing that this is a IIIF link and as such it should be added `/full/full/0/default.jpg` if directly put into an `` tag (otherwise it returns 30X to `info.json`)? Otherwise the new article JSON may break journal figure pages?
/cc @thewilkybarkid

thewilkybarkid: Old property is `uri`, new `image` property is a IIIF endpoint so when Journal is changed it will turn it into a full image URI.

giorgiosironi: :+1:

gnott: The image block url attribute will be the IIIF server value now as a result of this PR merge. Is this ok at this stage, or is this PR on the RAML schema related and should be adapted for it? https://github.com/elifesciences/api-raml/pull/122

PR #155, opened 2017-03-28
PLEASE DO NOT MERGE - checking validation of IIIF image values

gnott:


src/conf.py
```diff
@@ -125,6 +125,7 @@ def json_load(path):
 
 # NOTE: do not move to /tmp
```
giorgiosironi: @lsh-0 left written not to move these caches to `/tmp`, but we definitely need a more stable location as the project folder will be often empty during builds (it's a new Jenkins workspace). Only on `lax--*` servers these will work well, but in the `elife-libraries--*` one that builds PR and the `develop` branch the cache will always be empty. We could have a configuration in the .cfg file to choose the root directory for caches?

gnott: I added "cache_path" to start where you could specify a relative path such as ../  That can be adjusted if it should get its own absolute path format?

giorgiosironi: yes, we will do so through the formula or the build, thanks

lsh-0: yes, `/tmp` is emptied on ec2 `stop` and `start`, undermining the point of a cache and spamming Glencoe unnecessarily.  

perhaps a `/var/cached/` in builder-base ... ?

giorgiosironi: `/var/cache` is standard Unix so `/var/cache/apps` or similar?

giorgiosironi:
Was pointing to `dev--iiif.elifesciences.org`, which does not exist, hence the fallback to `"width":1, "height": 1`. Added a commit to point it in the right direction, now looking at the cache folder since it needs to work in many disparate situations.

giorgiosironi:
Added a commit that tries to use `/home/elife` as the default cache path, this is the build server:
```
elife@powerful--libraries:~$ ls -l
total 1084
lrwxrwxrwx 1 elife elife      42 Dec  5 10:30 elife-article-xml -> /ext/cached-repositories/elife-article-xml
-rw-r--r-- 1 elife elife    5120 Mar 31 10:28 glencoe-cache.sqlite3
-rw-r--r-- 1 elife elife 1095680 Mar 31 10:28 iiif-cache.sqlite3
```
I'll check lax-formula to see how it will be set there.

giorgiosironi:
@gnott opened https://github.com/elifesciences/lax-formula/pull/7 which I merged

giorgiosironi:
Now https://ci--alfred.elifesciences.org/job/pull-requests-projects/job/bot-lax-adaptor/job/PR-155/lastBuild/console is running, this will probably take a long time...

giorgiosironi:
I'll trigger a new build to see if it is the same ones that fail. If so, we have a data compatibility problem. If not, we have a stability problem.

src/conf.py
giorgiosironi: we went for `/home/elife` for now

giorgiosironi:
They are the same, which suggests a permanent failure on some images here:
```
17:05:01 INFO - 2017-03-31 15:45:28,849 - elife-13232-v2.xml -> elife-13232-v2.xml.json => failed (unhandled status code from IIIF: 502)
17:05:01 INFO - 2017-03-31 15:45:40,174 - elife-12997-v2.xml -> elife-12997-v2.xml.json => failed (unhandled status code from IIIF: 502)
17:05:01 INFO - 2017-03-31 15:57:31,119 - elife-04805-v2.xml -> elife-04805-v2.xml.json => failed (unhandled status code from IIIF: 502)
17:05:01 INFO - 2017-03-31 16:03:09,468 - elife-01370-v1.xml -> elife-01370-v1.xml.json => failed (unhandled status code from IIIF: 502)
```

src/conf.py
lsh-0: no worries. I'll look at setting up a /var/cache/appname/ state in the future

giorgiosironi:
There is (possibly) a missing `RGBX` key in one of Loris's Python modules. I'll try opening a pull request to them.

giorgiosironi:
https://github.com/loris-imageserver/loris/pull/286 but we are also deploying our local fork to cherry-pick this bugfix (and others which we may do in the future).

giorgiosironi:
:muscle:

PR #156, opened 2017-03-30
Differentiating between CDN_IIIF and IIIF

giorgiosironi:
/cc @gnott


gnott:
:+1: thanks @giorgiosironi

PR #159, opened 2017-04-06
Kitchen sink 00666 used in end2end fails validation

giorgiosironi:
We do not test it here, but we should. ~~If the version is incorrect, we need to update the version and then retest it.~~


giorgiosironi:
This version is out of date, will get a newer one in.

PR #164, opened 2017-04-11
Updated api-raml to: 2017-03-27 09:14:59 +0000

elife-alfred-user:
I have run http://ci--alfred.elifesciences.org/job/dependencies-bot-lax-adaptor-update-api-raml/2/ which resulted in this pull request.


giorgiosironi:
This is the RAML that validates IIIF figures and videos. It seems that all of bot-lax-adaptor's output already validates with that, as expected.

PR #165, opened 2017-04-11
IIIF image width and height do not use a default of 1.

gnott:


gnott:
This passes validation, although the test cases I had to put in None for width and height, which is not really valid. We could mock the IIIF server response to insert some non-None values (maybe).

Is this good policy to not set a default of 1 pixel and use None (which will fail validation?). So far it doesn't cause any validation problems.

giorgiosironi:
Important: we cannot merge until https://elifesciences.atlassian.net/browse/ELPP-2495 is done as this will work on backfills but not for new articles as the images are not available through IIIF yeat at the time bot-lax-adaptor is ingesting the XML.

This PR doesn't fail validation because all published articles in elife-article-xml have valid images served through IIIF now, good news!

As for the error, I would go for an hard exception on generation rather than waiting for validation to catch this with a huge and complex validation error message? We can give better feedback to the user making explicit with URL on the IIIF server is not loading correctly. Besides, the bot will execute the VerifyImageServer activity before ingestion to avoid the bot-lax-adaptor running on missing IIIF images.

giorgiosironi:
For the test cases: maybe we could point to https://continuumtest--iiif.elifesciences.org/ in unit tests instead of production, since it should have the kitchen sink?

giorgiosironi:
https://github.com/elifesciences/bot-lax-adaptor/pull/165 is in review now, and deployed in production.
As specified in https://github.com/elifesciences/elife-bot/blob/develop/workflow/workflow_IngestArticleZip.py#L126-L137 `VerifyImageServer` should protected this code from failing, let's see if this holds true.

giorgiosironi:
Failed when in `develop`, how could this pass on the branch?
https://ci--alfred.elifesciences.org/job/library-bot-lax-adaptor/239/consoleFull

giorgiosironi:
We'll have to use continuumtest when generating the kitchen sink in `kitchen-sinks.sh`, the only place where there is a real IIIF server containing its images (if it really contains all its images).

gnott:
It might not be reliable always depending on what is in continuumtest. I think mocking the IIIF server response might be a good idea for the kitchen sink.

giorgiosironi:
https://github.com/elifesciences/bot-lax-adaptor/blob/develop/kitchen-sinks.sh is outside of the Python process, so it can't mock requests; however we could point it to the `ci` IIIF server so that it could be more reliable.

gnott:
Ah yes, I guess the ci server will always have the kitchen sink images? As long as the tests pass :+1:

giorgiosironi:
Done a quick experiment, it's even worse as the image it's looking for are:
```
https://s3.amazonaws.com/continuumtest-elife-published/articles/00666/elife-00666-fig1.tif
```
without the `-vX` suffix. These will not exist in any IIIF server, and even if we get them in `ci` we would have to place them manually inside a bucket and maintaining them.
The only thing I can do to keep around the validation of the kitchen sink in this build (rather than removing it) is to have a test setting that keeps the fake width. Be right back with a proposal.

giorgiosironi:
Let's see if https://github.com/elifesciences/bot-lax-adaptor/pull/181 works

PR #175, opened 2017-04-19
Do not include subjects if the list is empty.

gnott:


gnott:
All validates and no harm so I will merge.

PR #176, opened 2017-04-20
Log error responses

giorgiosironi:
The loggers set up on stderr and on adaptor.log ignore the extra parameter, hence error responses are not logged. By putting the responses inside the formatted message they become accessible.

Moreover, the message in http_download does not need to take into account the context of the file being an XML, as this is redundant with the message set in the `except` clause that catches it

### DIFF
--- a/src/adaptor.py
+++ b/src/adaptor.py
@@ -114,7 +114,7 @@ def http_download(location):
             cred = S3Auth(*cred)
     resp = requests.get(location, auth=cred)
     if resp.status_code != 200:
-        raise RuntimeError("failed to download xml from location %r, got response code: %s" % (location, resp.status_code))
+        raise RuntimeError("failed to load %r, got response code: %s\n%s" % (location, resp.status_code, resp.content))
     resp.encoding = 'utf-8'
     return resp.text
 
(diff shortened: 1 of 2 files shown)


src/adaptor.py
```diff
@@ -157,7 +157,7 @@ def mkresponse(status, message, request={}, **kwargs):
         INGESTED: logging.DEBUG,
         PUBLISHED: logging.DEBUG
     }
-    LOG.log(levels[packet["status"]], "%s response", packet['status'], extra=context)
```
lsh-0: did you mean to leave off the `extra=` in the log line?

lsh-0:
> The loggers set up on stderr and on adaptor.log ignore the extra parameter
rather than rolling into into the message, I'd look at modifying the format lines for those logs to include the 'extra' context. good catch.

giorgiosironi:
Feel free to amend, end result should be that on the stderr log we see the extra parameters (usually this is captured in `/var/log/upstart/bot-lax-adaptor-1.log`)

lsh-0:
it's a little hacky because the dict passed to `extra` is merged into the Record object they're passing around behind the scenes, so what this does is make an 'extra' field available to the format line whose value is a json encoded map of all of the unidentified fields.

there's also a wee bit at the bottom of conf.py you can use to see the result for yourself.

feel free to modify the format line and merge if you're happy

PR #183, opened 2017-04-27
Slowly removing old image schema

giorgiosironi:
- Removing alt and uri fields (These have been substituted by the ones inside the 'image' dictionary)
- Removing filename outside of misc/image type (There never was a filename here: only inside the 'source' dictionary https://github.com/elifesciences/api-raml/pull/122/files#diff-bcb896b4a098e003f412ac6257490386R1456)
- Correct comment about assumed file extension

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -340,7 +340,6 @@ def expand_uris(msid, data):
             element['uri'] = fixed
             return element
         # normal case: cdn link
-        element["filename"] = os.path.basename(element["uri"]) # basename here redundant?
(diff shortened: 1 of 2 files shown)


giorgiosironi:
(waiting for reviews from all)

giorgiosironi:
Once this is in, I'm totally in favor of removing code that is generating fields not being used (or being actually removed). Wanted to pin down a good output first so that then we can refactor the internals.

PR #186, opened 2017-05-02
Utility script for backfilling single article

giorgiosironi:
Doing it through src/backfill.py seems the easiest way

### DIFF
--- a/backfill-single.sh
+++ b/backfill-single.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -e
+
+if [ "$#" != 3 ]; then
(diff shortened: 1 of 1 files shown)


lsh-0:
you may be able to avoid having to specify the `msid` and `version`, the code in backfill.py very quickly hits this: https://github.com/elifesciences/bot-lax-adaptor/blob/develop/src/fs_adaptor.py#L14

giorgiosironi:
Not sure I want it to be that magic, we may have strangely named files when we try to fix a problematic article.

PR #187, opened 2017-05-02
Using cdn.elifesciences.org

giorgiosironi:
This is only for production however, because CDN1 does not exist for end2end and continuumtest. Creating it and coming back here, but at least we can try to build the corpus from the CDN1.

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -129,7 +129,7 @@ RESPONSE_SCHEMA = json_load('response-schema.json')
 CDN1 = 'cdn.elifesciences.org/articles/%(padded-msid)s/%(fname)s'
 CDN2 = 'publishing-cdn.elifesciences.org/%(padded-msid)s/%(fname)s'
 
-DEFAULT_CDN = CDN1 if False else CDN2 # 'False' until we finish switching to a single CDN :(
+DEFAULT_CDN = CDN1
 CDNS_BY_ENV = {
     'end2end': 'end2end-' + CDN2,
     'continuumtest': 'continuumtest-' + CDN2,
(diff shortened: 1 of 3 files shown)


giorgiosironi:
In an attempt to get beta.elifesciences.org working again I'll merge this, but @lsh-0 @gnott note here if something is amiss. A follow-up would be to use (a different but identical instance of) the new CDN in `end2end` and `continuumtest` too.

gnott:
:+1:  I hope it works as planned!

lsh-0:
oh wow - *completely* overlooked in that CDN ticket. shit. I visited beta.elifesciences.org after turning the cdn off and nothing seemed amiss - in fact, I was really impressed with how far it had come and how polished it was looking from last time I visited. Nothing seemed broken, articles rendered fine. Excellent catch

lsh-0:
> (a different but identical instance of) the new CDN in end2end and continuumtest too.

this is an interesting point. getting rid of publishing-cdn was partly about understanding why we had a second cdn, second data store, who was using what and why and then consolidating them under the single cdn. with the switch to 2.0 we're already doing DNS gymnastics with platform.sh - is there an opportunity here to use a cloudformation-provided cdn and still retain the `cdn.elifesciences.org` address?

giorgiosironi:
As you have seen and approved on https://github.com/elifesciences/builder/pull/133#pullrequestreview-35920900 the CDNs equivalent to `cdn.elifesciences.org` for `end2end` and `continuumtest` are now provided through CloudFormation. There is nothing stopping us from doing the same for `prod` apart from downtime, because the minimal steps are:
1. creating the new CDN with Aliases value `prod--cdn.elifesciences.org` (~60 mins but it's not downtime)
2. deleting the CNAME `cdn.elifesciences.org
3. deleting the old CDN, freeing up the Aliases value `cdn.elifesciences.org` (~60 mins)
4. updating the new CDN, adding the Aliases value `cdn.elifesciences.org` (~20 mins)
5. adding the CNAME `cdn.elifesciences.org` pointing to the new CDN

Step 3 can probably be reduced (usually to ~20 mins) by removing the Aliases value rather than deleting the whole CDN, which is also safer for rollbacks.

PR #191, opened 2017-05-08
Updated elife-article-xml to: 2017-05-06 12:38:06 +0100

elife-alfred-user:
I have run http://ci--alfred.elifesciences.org/job/dependencies-bot-lax-adaptor-update-elife-article-xml/12/ which resulted in this pull request.


lsh-0:
I didn't propose any fixes at the time. Is the problem solely within  bot-lax-adaptor? is it the multiprocessing that's causing this? or is there another thing also interacting with it? if it's just bot-lax, I would switch to another backend: https://requests-cache.readthedocs.io/en/latest/user_guide.html#persistence

I'll open a ticket and assign it to me so this doesn't get lost

giorgiosironi:
I think it is the multiprocessing, causing too much concurrency on the Sqlite database. I'll see if #194 gets built.

PR #194, opened 2017-05-09
attempt to avoid sqlite cache db lock problem

lsh-0:


giorgiosironi:
It builds correctly, no effect on performance (everything should be cached so this may vary when we actually write to it). Let's try it and see if the problem happens again, it's very rare anyway.

PR #201, opened 2017-05-15
Avoid IIIF prefix on prod

thewilkybarkid:


src/conf.py
```diff
@@ -135,10 +135,13 @@ def json_load(path):
 }
 CDN = 'https://' + CDNS_BY_ENV.get(ENV, DEFAULT_CDN)
 
-if ENV in ['prod', 'continuumtest', 'end2end']:
+if ENV == 'prod':
     # used for generating public links
-    CDN_IIIF = 'https://' + ENV + '--cdn-iiif.elifesciences.org/lax:%(padded-msid)s/%(fname)s'
+    CDN_IIIF = 'https://iiif.elifesciences.org/lax:%(padded-msid)s/%(fname)s'
```
giorgiosironi: for reference on where this domain comes from: https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L900

thewilkybarkid:
We'll need to backfill for the figures change, no rush.

PR #213, opened 2017-05-19
Add figuresPdf depending on the file names of graphics in the XML.

gnott:
An alternative method to determine whether there should be a figures PDF.


giorgiosironi:
Does the kitchen sink have some of these interesting tags so that we can check its generated output?

Anyway, let's hold this, I'll ask at stand up for which option we want to choose for figures PDF (XML or file check).

giorgiosironi:
It is now confirmed that the other method (lookup through HTTP call on prod-elife-published) is more reliable and should be used instead.

PR #214, opened 2017-05-22
Bugfix: recover from sqlite lock error

lsh-0:
replaced requests.get with utils.requests_get that will try three times to fetch a result if a sqlite3 operational error occurs.


src/utils.py
```diff
@@ -157,6 +161,25 @@ def partial_match(pattern, actual):
 #
 #
 
+def call_n_times(fn, protect_from, num_attempts=3):
+    "if after calling `num_attempts` it fails to return a value, it will return None"
```
giorgiosironi: `None` does not seem a particularly informative value in case of prolonged failures, but at least whatever is calling will fail too e.g. when it tries to access a field (and we have longs)

lsh-0: yeah. I actually struggled a bit with this because of the semantics. 
`num_attempts` would presume the function is called `num_attempts` times, but it feels like it should fail naturally *after* n times, so: "call n times then fail". The over-engineered solution that I scrapped supported this with being called n times with the last time not being protected. easy enough to add one more unprotected call in `requests_get`

PR #205, opened 2017-05-17
PLEASE DO NOT MERGE - Include figuresPdf and refactor requests_cache configuration …

gnott:
…to be universal.

An experiment.

PR #221, opened 2017-05-25
Request directly info.json

giorgiosironi:
If we get redirects from the image URL to the info.json, we are then
incapable of clearing the cache for those redirects as some of the
characters are encoded in a non-trivial way (see
https://github.com/elifesciences/bot-lax-adaptor/pull/205#issuecomment-303955552)

By requesting directly info.json, we avoid redirects and only this URL
should be cached (and pruned from the cache on errors).

This will be a one-off performance hit as we have to repopulate the
cache.


gnott:
:+1:

PR #235, opened 2017-06-02
feature: new backfill 

lsh-0:
This PR is tied to the similarly named one in lax: https://github.com/elifesciences/lax/pull/213

* replaces `backfill-published.sh` and `backfill-unpublished.sh` with `backfill.sh`
* uses Lax as the source of truth for all known articles, published and unpublished
* downloads any known articles from S3 that are missing from the article xml repo (master branch)
* renames `backfill.py` to `adhoc_backfill.py`. this script sends small numbers of articles through the `fs_adaptor.py`
* backfill happens in `/tmp` by default, or `/ext` if it exists.


backfill.sh
```diff
@@ -0,0 +1,140 @@
...
+    exit 1
+}
+
+srcdir=$(pwd)
+errcho(){ >&2 echo $@; }
+
+# where articles will be linked to/downloaded for backfill
+defaultrunpath="run-$(date +'%Y%m%d%H%M%S')"
+runpath=${1:-$defaultrunpath}
+runpath=$(realpath $runpath) # realpath needs to be installed :(
```
giorgiosironi: lax-formula? builder-base-formula?

backfill.sh
```diff
@@ -0,0 +1,140 @@
...
+mkdir -p "$runpath"
+
+shift # download-api-raml.sh takes an arg :P
```
giorgiosironi: what is the purpose of this? I see no further interaction with parameters

backfill.sh
```diff
@@ -0,0 +1,140 @@
...
+
+# where to find xml on the fs
+xmlrepodir="$srcdir/article-xml/articles"
+
+# where to download unpublished xml to
+unpubxmldir="$srcdir/unpub-article-xml"
+mkdir -p "$unpubxmldir" # (create if necessary)
+
+# where generated article-json will be stored
+ajsondir="$runpath/ajson"
```
giorgiosironi: currently the `article-json` folder is symlinked to `/ext` because of its size. Maybe `$runpath/ajson` could be created as `./article-json/run-2017.../` and symlinked in `$runpath` instead so that every new run ends up on the external volume. Especially since new runs won't overwrite older ones, so we need to keep space under control.

backfill.sh
```diff
@@ -0,0 +1,140 @@
...
+    
+    # ll: /home/user/bot-lax/unpub-article-xml/elife-00003-v1.xml
+    xmlunpubpath="$unpubxmldir/$fname"
+    
+    # we look in both places for xml and if it's in neither, we download it
+
+    if [ ! -f $xmlpath ] && [ ! -f $xmlunpubpath ]; then
+        # lax doesn't know where the remote location and it's not on the fs
+        if [ $remotepath = "no-location-stored" ]; then
+            errcho "$fname not found, skipping"
```
giorgiosironi: ok, we skip it but we have a log

backfill.sh
```diff
@@ -0,0 +1,140 @@
...
+    fi
+
+    #errcho "linking $fname"
+
+    # link it in to the run dir
+    if [ -f $xmlpath ]; then
+        ln -sfT $xmlpath $fname
+    else
+        ln -sfT $xmlunpubpath $fname
+    fi
```
giorgiosironi: ok, so if the `article-xml` repository has an XML we take that one, otherwise it means it's either a) unpublished or b) not yet in the repo, in both cases downloading it from the remote path is the best choice.

This doesn't cater for silent corrections however, if the sha of `article-xml` does not have them (because it's slightly out of date) the old version in there will be used instead of the one silently corrected. This has to be solved as it is dangerous. Perhaps we should switch to the `master` of `article-xml` for the backfill? We won't have the guarantee that all articles are definitely valid with this code as the more recent ones haven't been tested, but they got published so they were valid before - it's a small price to pay.

src/generate_article_json.py
```diff
@@ -24,19 +28,24 @@ def render(path):
...
+    parser.add_argument('output-dir', nargs='?', default=conf.JSON_DIR)
+
+    args = vars(parser.parse_args())
+    indir, outdir = [os.path.abspath(args[key]) for key in ['xml-dir', 'output-dir']]
+
+    ensure(os.path.exists(indir), "the path %r doesn't exist" % indir)
+    ensure(os.path.exists(outdir), "the path %r doesn't exist" % outdir)
+
+    main(indir, outdir)
+    exit(0)
```
giorgiosironi: do you actually need to `exit(0)`, isn't it the default?

src/validate_article_json.py
```diff
@@ -12,48 +11,50 @@
...
 
-WINDOWS = platform.system().lower() == 'windows'
+def copyfile(src, dest):
+    if os.path.exists(dest):
+        os.unlink(dest)
+    return shutil.copyfile(src, dest)
 
-def job(path):
-    strbuffer = StringIO()
+if platform.system().lower() == 'windows':
```
giorgiosironi: I don't think we really support `windows`? :)

backfill.sh
lsh-0: ah yes, some distros have it installed by default, some not. I was going to see what Ubuntu had by default (it doesn't). I think there are clever bash-isms to resolve symlinks and get the full absolute path , but they're so arcane ... I'll update the builder-base now to have it installed by default.

backfill.sh
lsh-0: big badda boom: https://github.com/elifesciences/bot-lax-adaptor/blob/develop/download-api-raml.sh#L8

took me a while to find it as well. I didn't think subsequently sourced scripts would have access to params passed to parent, but it turns out they do. what `shift` does is pop the first arg off of `$@` and return it. in this case I don't capture the return value and downstream scripts don't see the first arg.

backfill.sh
lsh-0: as I was picking the code apart I could see we're actually using twice the amount of space than we need to. way back in the beginning we were 'patching' the generated article json to fill in the missing gaps and getting it passing validation so lax wouldn't reject it, lax would then fix up anything it needed to under the hood. 

the `generate-article-json` stage creates the article-json without the patched values and then the validate stage patches this article-json before validating it.

the backfill process is generating the article-json, then validating it and storing the patched version in `$ajsondir/patched` and then sending the patched version to lax.

I could halve the space requirements by adding the patching to `main.py` in the post-process step after scraping and having validate.py deal with just `valid` and `invalid`.

It wouldn't remove the need for an external disk - the article xml will grow over time - but it would make things more efficient and less unwieldly. I'm sure those inodes would thank me if they could as well.

backfill.sh
lsh-0: urgh. if we could have switched to master branch of article-xml for the backfill I might have avoided the rewrite. that said, this rewrite feels like the right thing. it's removing a tonne of now-unnecessary code in bot-lax and lax.

... I'm trying to figure out what else is using the article-xml repo in lax or bot-lax.  in bot-lax it's being used in `web.py` which was a quick and dirty web interface I threw together that could be torn out/disabled/made more flexible, and it's used as the default in the `backfill.py` script. no mentions of it in lax.

should we just dispense with pinning it altogether?

if we can't, then I might be able to add support to lax to record silent corrects (I think I already do ...?) and add another response besides 'no-location', like 'silent-correction' for articles that have been silently corrected in the last N days.

src/generate_article_json.py
lsh-0: yeah. I saw that, thought the same thought and was too lazy to `ctrl-k` it. removing now.

lsh-0: fixed. in two other places as well

src/validate_article_json.py
lsh-0: @gnott , do you still bulk generate article json?

backfill.sh
lsh-0: I'll do a branch of this branch with the proposed changes.

src/validate_article_json.py
gnott: I have not for a long time, no. I don't think I run this file (knowingly).

lsh-0: thanks @gnott . I'll just slice it out then.

backfill.sh
giorgiosironi: then we should use `sh` instead of `source` to create a subprocess rather than include the other script? overhead is minimal and isolation should prevent these problems

backfill.sh
giorgiosironi: - to pin it for tests is useful, because when you run the same code you have the same result (and it takes tens of minutes to have a result)
- to pin it for production is desirable but marginally less useful as there are external conditions (like the state stored in lax) that can produce a different result anyway
If lax can say what has been silently corrected it is probably a useful information not just for this task but in general as a structured log of changes? Every silent correction is a git force push that breaks the update model. I'd say go for unpinning it on the backfill if necessary, but not unpinning it in tests for reproducibility.

Big *but*: I have to check what the code does now, but if we favor the Lax-stored XML location rather than the article-xml, doesn't that solve silent corrections because during one the Lax-stored XML location is updated?

giorgiosironi: Answer: we could favor the Lax-stored XML location, but it would be very slow to download all of them. For me unpinning the article-xml repo on backfill + using the remote S3 locations for unpublished articles covers all cases and is sufficient for now.

backfill.sh
lsh-0: do you mean `. some-script.sh` ? or switch from bash to sh ? because I tried the former and it doesn't work

backfill.sh
lsh-0: https://github.com/elifesciences/bot-lax-adaptor/pull/239

backfill.sh
lsh-0: I'll do that then

lsh-0: lax has the beginnings of an `ArticleEvent` table that I was hoping to shift the EJP events into as they're really not well placed currently in the `Article` table: https://github.com/elifesciences/lax/blob/develop/src/publisher/models.py#L315

it's a little unloved right now but it's something I've been wanting to do for a while.

backfill.sh
giorgiosironi: `sh some-script.sh`? or just `./some-script.sh`?

backfill.sh
lsh-0: :+1: merged in

lsh-0:
ok, fixes are in. I'll give it a go on end2end once it's finished testing

lsh-0:
tested on an adhoc instance. it's good to go in my opinion. @giorgiosironi - could you do the honours please?

giorgiosironi:
:rocket:

PR #239, opened 2017-06-07
Feat: new backfill, halve output

lsh-0:
the current backfill uses disk space for:

* article xml
* article json
* patched article json

the patched article json is then sent to lax


lsh-0:
this felt like too large of a change during the backfill rewrite to consider, but looking at the changes now I don't it's that much of deal

PR #243, opened 2017-06-13
No maximum version

giorgiosironi:
Conceptually there are no limits to version numbers in the api-raml, so we may remove this assumption. Originally uncovered by Graham testing the 10th version of the kitchen sink article (00666).


gnott:
I'm ok with it :+1:

lsh-0:
ah - well done. I completely forgot I wrote those.

PR #247, opened 2017-06-21
Use latest parser, new copyright_holder_json, add full stop to et al.

gnott:


giorgiosironi:
It looks like this needs to be merged? Not sure if it's ready.

PR #257, opened 2017-07-04
Feature: validate

lsh-0:
adds an api to bot-lax that replaces the old web interface I cobbled together once. 

it's purpose is to allow ad-hoc uploads of xml with replacement values for fields and ensure it gets scraped, validated and ingested into lax (as a dry run, no modifications to lax).

lax and bot-lax now make an extra effort to return responses that can be understood by a human, with a 'trace' attribute in error responses for stack traces and validation errors for those who must debug the error.


giorgiosironi:
This build stopped using the request cache, ddosing iiif--prod by loading all possible info.json; at least build 4 and 5 seem to have triggered the problem (see it-admin emails).

giorgiosironi:
Didn't probably stop using the request cache, it's rather writing a different format to it which cause all other builds to fail when they cannot unserialize the values:
```
  File "/ext/jenkins-libraries-runner/workspace/or_PR-260-RRS2HHGL72HVBTUQAKEJ/venv/local/lib/python2.7/site-packages/requests_cache/backends/storage/dbdict.py", line 163, in __getitem__

    return pickle.loads(bytes(super(DbPickleDict, self).__getitem__(key)))

ImportError: No module named urllib3._collections
```

giorgiosironi:
`requirements.txt` changes should be taken away and done in another PR, they seem to have a large impact then.

giorgiosironi:
Also, the name of the cache files could be changed during the upgrade in order to avoid conflicts between the two versions - still requires hitting all the info.json again but lets different branches run continuously.

lsh-0:
I've downgraded the requirements (not pushed), but I'm worried about the testing environment potentially bringing down production.

from the src/conf.py file:
```
elif ENV in ['continuumtest', 'end2end']:
    CDN_IIIF = 'https://' + ENV + '--cdn-iiif.elifesciences.org/lax:%(padded-msid)s/%(fname)s'
    IIIF = 'https://' + ENV + '--iiif.elifesciences.org/lax:%(padded-msid)s/%(fname)s/info.json'
```

how is it touching the iiif-prod server?

giorgiosironi:
The tests run in a `ci` environment, hitting the production servers because they are the ones pointing to the full corpus. This hasn't been a problem because of the cache, until the cache was invalidated.

lsh-0:
finally completed the gauntlet!

elife.cfg
```diff
@@ -1,6 +1,7 @@
 [general]
 env: dev
+#upload_path uploads
+#cache_path: cache
 
 [lax]
 location: /srv/lax/
-cache_path: /home/elife
```
giorgiosironi: Also `/var/cache/` or similar would be fine: the important thing is for the cache to be kept by default outside of the project folder, which is created from scratch for every new pull request

schema/api.yaml
```diff
@@ -0,0 +1,111 @@
...
+        description: the final state of the article supplied
+      ajson:
+        type: object
+        description: the final article-json, including overrides
+      override:
+        type: object
+        description: the overrides provided
+
+paths:
+  /xml:
```
giorgiosironi: @thewilkybarkid can confirm paths and field names conform to our API style (this is not part of api-raml but it pays to be consistent)

schema/request-schema.json
```diff
@@ -1,35 +1,39 @@
...
-		}
-	},
-	"required": ["action", "location", "id", "version", "token"]
+    "description": "lax ingest+publish request schema",
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "id": "https://lax.elifesciences.org/ingest-publish-request-schema.json",
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+    "action": {
```
giorgiosironi: this reformatting seems to have moved to the wrong indentation the contents of `"properties"`

src/adaptor.py
```diff
@@ -190,8 +207,8 @@ def handler(json_request, outgoing):
     # we have a valid request :)
     LOG.info("valid request")
 
-    params = subdict(request, ['action', 'id', 'token', 'version'])
-    params['force'] = request.get('force') # optional value
+    params = subdict(request, ['action', 'id', 'token', 'version', 'force', 'validate-only'])
+    params = renkeys(params, [('validate-only', 'dry_run')])
```
giorgiosironi: `validate-only` and `dry_run` are using two different conventions (`-` and `_`), which one are we targeting? `api-raml` usually goes for `camelCase`

src/utils.py
```diff
@@ -38,9 +38,16 @@ def ensure(assertion, msg, *args):
     if not assertion:
         raise AssertionError(msg % args)
 
+def writable_dir(path):
```
giorgiosironi: good check, failing fast

src/api.py
```diff
@@ -0,0 +1,239 @@
...
+        orig_path = super(BotLaxResolver, self).resolve_operation_id_using_rest_semantics(operation)
+        bits = orig_path.split('.') # ll ['api', 'xml', 'search']
+
+        module = bits[0] # ll: 'api'
+        # the funname resolution is wonky, so discard it
+        method = bits[-1] # ll: 'search'
+
+        path = operation.path # ll: /article-json/validation/{filename}
+        # we want something like:
+        # /article-json/validation/{filename} => article_json_validation
```
giorgiosironi: which of the two orders is the correct one?

giorgiosironi: (why supporting two paths leading to the same method when only one is needed?)

src/api.py
```diff
@@ -0,0 +1,239 @@
...
+            'code': 'invalid-article-json',
+            'message': 'the generated article-json failed validation',
+            'trace': str(err), # todo: any good?
+        }, 400
+    except Exception as err:
+        sio = StringIO()
+        traceback.print_exc(file=sio)
+        return {
+            'status': conf.ERROR,
+            'code': 'error-validating-article-json',
```
giorgiosironi: are these `code`s documented somewhere? The Swagger file could list them? Clients will be keen to know the full list

src/api.py
```diff
@@ -0,0 +1,239 @@
...
+        # 'code', 'message' and 'trace' are not returned by lax on success, just 'status'
+        api_resp['ajson'] = json.loads(article_json)['article']
+        api_resp['override'] = override
+        return api_resp, 200
+
+    except Exception as err:
+        # lax returned something indecipherable
+        return {
+            'status': conf.ERROR,
+            'code': 'lax-is-borked',
```
giorgiosironi: not sure if client needs to know about `lax`, from their point of view there is some API accepting article XMLs.  Can the `code` be made more generic?

src/conf.py
```diff
@@ -71,84 +102,78 @@ def multiprocess_log(filename, name=__name__):
...
 
-REQUEST_SCHEMA = json_load('request-schema.json')
-RESPONSE_SCHEMA = json_load('response-schema.json')
+REQUEST_SCHEMA = load('request-schema.json')
+RESPONSE_SCHEMA = load('response-schema.json')
+
+API_SCHEMA = load('api.yaml')
+
+# can be overriden when creating an app
+API_UPLOAD_FOLDER = join(PROJECT_DIR, "uploads")
```
giorgiosironi: this is in the formula right?

elife.cfg
lsh-0: I'd prefer to keep ci/deployment details out of the application, I don't even like environment details being present, so what if I updated project_tests.sh to emit a better `app.cfg` file?

schema/request-schema.json
lsh-0: I've reverted indentation now

src/adaptor.py
lsh-0: those are function parameters and not part of the json response. here, the request is converted into something more suitable for internal use, renaming `validate-only`, to `dry_run`. in retrospect I should have just used 'dry-run' across the board. It evolved from being a discreet action like ingest or publish to being an option.

src/api.py
lsh-0: the paths changed a bit during dev, it's possibly an old comment that has become confusing. I'll clean it up

lsh-0: it's not an old comment, it's just illustrating what happens when a path contains path parameters and gets flattened. the final path has the method (`get`, `post`, etc) prefixed to it so that should help to differentiate, but if our future selves experience a conflict in function name resolution then we're probably doing it wrong.

src/api.py
lsh-0: codes are not documented anywhere like that are in lax. they probably should be and an enum in the schema would be helpful.

src/api.py
lsh-0: yes, and the `borked` message is an oversight as well ;)

src/conf.py
```diff
@@ -71,84 +102,78 @@ def multiprocess_log(filename, name=__name__):
...
 PATH_TO_LAX = cfg('lax.location')
 
-CACHE_PATH = cfg('lax.cache_path')
+# TODO: re-enable once changes to lax-formula are deployed
+CACHE_PATH = join(PROJECT_DIR, 'cache')
+# if ENV != DEV:
+#    CACHE_PATH = cfg('general.cache_path', CACHE_PATH)
+
+# TODO: remove once changes to lax-formula are deployed
+CACHE_PATH = cfg('lax.cache_path', CACHE_PATH)
```
lsh-0: this is incorrect, it shouldn't be reading from the lax section but the general section. I'll fix that up

src/conf.py
lsh-0: yes indeed: https://github.com/elifesciences/lax-formula/blob/feat-VALIDATE/salt/lax/bot-lax-adaptor.sls#L98

and git has that folder created with a placeholder inside.

this state should be safe until the formula updates introduce the bot-lax uwsgi server, at which point the uploads and cache and log directories need to be writable by `www-data` and `elife`. I managed to remove the elife-tools `test.log` from the equation. 

the bot-lax app.cfg file: https://github.com/elifesciences/lax-formula/blob/feat-VALIDATE/salt/lax/config/opt-bot-lax-adaptor-app.cfg

I'll re-enable the upload folder to read from the cfg file, this shouldn't affect anything until the formula kicks in. the log folder and cache folder might actually be problems now that I look at them. it depends on them existing with the right permissions. might need to split the formula changes into two sets

src/conf.py
lsh-0: ah, I see what I was doing now. I've added `cache_path: /home/elife` back into the cfg file with a comment. we still need to stitch CI up with it's own app.cfg and a path to a shared cache file in the future, though.

elife.cfg
giorgiosironi: That's equivalent, good solution too - and project_tests.sh is specific to `ci`

src/api.py
lsh-0: codes are now in conf.py and documented in swagger: 
screenshot at 2017-07-17 13-10-55

lsh-0:
ok, I think I've addressed your points, I'm going to merge this into develop now and cross my fingers and toes.

giorgiosironi:
:+1: for the merge

PR #259, opened 2017-07-05
ELPP-2961 expand_location botching URLs

giorgiosironi:


giorgiosironi:
`develop` is broken, have to fix that first.

lsh-0:
ah - is this the culprit causing broken locs? I arrived at this second of code one night trying to debug this and left off for the evening, never to return to it. I wrote the report task 'fix broken locs' instead, which was just throwing code at the problem.

giorgiosironi:
It is, added the Jira ticket in the title.

PR #260, opened 2017-07-05
Reset virtualenv on build

giorgiosironi:
https://ci--alfred.elifesciences.org/job/library-bot-lax-adaptor/298/console failed for a leftover virtualenv, so in builds we can construct it from scratch for better reproducibility. end2end will still test the upgrade of an existing virtualenv

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -2,6 +2,8 @@
 
 set -e # everything must pass
 
+rm -rf venv/
+
 . install.sh
 
 . .lint.sh


giorgiosironi:
Not a problem of virtualenv, it's the cache trying to deserialize something that does not have a class anymore.

giorgiosironi:
Will have to delete the cache and retry, but avoiding impact on iiif--prod.

PR #261, opened 2017-07-07
upgraded deps, scrubbed

lsh-0:


lsh-0:
merging, doesn't really need review.

PR #264, opened 2017-07-11
Attribution field move

giorgiosironi:
- Switching elife-tools to branched version to check generation of attribution field
- Switching api-raml to branched version with new schema

See https://github.com/elifesciences/elife-tools/pull/217


giorgiosironi:
Closing in favor of #271

PR #276, opened 2017-07-24
Bugfix requests get silent failure

lsh-0:
should improve errors like these if nothing else: https://ci--alfred.elifesciences.org/job/library-bot-lax-adaptor/314/consoleFull


giorgiosironi:
I agree we should use that folder by default to keep the cache alive between builds of PRs. Not sure if this branch is up-to-date or should be closed as obsolete, given the conflict.

PR #322, opened 2017-10-04
Feature: fill missing image dimensions

lsh-0:


src/main.py
```diff
@@ -428,6 +428,26 @@ def check_authors(data):
...
+            and 'image' in element \
+            and 'size' in element['image']
+
+    def fix(element):
+        if not element['image']['size']['height'] or not element['image']['size']['width']:
+            element['image']['size']['height'] = 1
+            element['image']['size']['width'] = 1
+        return element
+
+    return visit(data, pred, fix)
```
giorgiosironi: Go go gadget Visitor pattern :+1:

src/api.py
```diff
@@ -131,7 +131,10 @@ def post_xml():
 
     # generate
     try:
-        article_json = scraper.main(path, {'override': override})
+        article_json = scraper.main(path, {
+            'override': override,
+            'fill-missing-image-dimensions': False
```
giorgiosironi: shouldn't this be passed as `True` in this `api` scenario then?

giorgiosironi: now that we only need an XML we should be able to call the API with a fixture (XML with figures) and validate the result?

lsh-0: > shouldn't this be passed as True in this api scenario then?
nuts, typo.

> now that we only need an XML we should be able to call the API with a fixture (XML with figures) and validate the result?

I'll ping you in the Jira ticket, the XML has been uploaded there. Add it to the continuum tests?

lsh-0: > Add it to the continuum tests?

or I could add it to the regular test suite...

lsh-0:
@giorgiosironi , happy with the test?

src/tests/test_api.py
```diff
@@ -360,3 +360,34 @@ def test_broken_glencoe_response(self):
...
+        xml_fixture = join(self.fixtures_dir, xml_fname)
+
+        expected_lax_resp = {
+            'status': conf.VALIDATED,
+            'override': {},
+            'ajson': base.load_ajson(xml_fixture + '.json')['article'],
+            'message': None # this should trigger an error when logged naively by api.py but doesn't...
+        }
+
+        # don't call lax
```
giorgiosironi: if it doesn't call lax, why we need to patch it?

giorgiosironi:
I think this has to wait on https://github.com/elifesciences/lax/pull/293 resolution anyway.

src/tests/test_api.py
lsh-0: because it will try to talk to lax if we don't. I thought it was clear, but what I'm testing here is what happens when `iiif.iiif_info` returns an empty result `{}` from the api. the response should be `200`, the response body should successfully validate as article json.

lsh-0:
that lax issue has been deployed now, merging this one in

PR #325, opened 2017-10-07
Updated elife-tools to: requirements.txt pinned to 1603ba144a32846623eadc16ab928825d3e4d8e4

elife-alfred-user:
I have run http://ci--alfred.elifesciences.org/job/dependencies-bot-lax-adaptor-update-elife-tools/45/ which resulted in this pull request.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ autopep8==1.3.2
 boto3==1.4.4
 coverage==4.1
 et3==1.5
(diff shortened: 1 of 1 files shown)


gnott:
This should have no effect on eLife articles, it parses issue values from article-meta which is only in non-eLife articles.

giorgiosironi:
There's a newer PR that makes this obsolete.

PR #334, opened 2017-10-18
lax now requires an envvar to be set during multiprocess backfill

lsh-0:


backfill.sh
```diff
@@ -172,6 +172,7 @@ echo > "$prjdir/validate.log"
 time python $prjdir/src/validate_article_json.py "$ajsondir"
 
 # call the lax 'ingest' command with a directory of valid article json
+MULTIPROCESSING=1
```
giorgiosironi: see https://github.com/elifesciences/lax/pull/296#pullrequestreview-70134309

PR #337, opened 2017-10-20
Bugfix fail early if pubdate not detected

lsh-0:


lsh-0:
are we talking about the `main` function or the whole `main.py` file?

I have no problems with 'busy-ness' and I don't expect fresh eyes to understand old code at a glance. The fact that you were able to get the gist of what is happening in there in only a short while sounds like a ringing endorsement of my code and your capability.

`main.py` or the 'scraper' as I sometimes import it as, has the fun job of building a very large and complicated data structure. Thankfully `main.py` file is well separated with the majority of the wrangling happening in elife-tools: https://github.com/elifesciences/elife-tools/blob/develop/elifetools/parseJATS.py

the scraper has two sections: building the body of the data structure and post-processing the built data structure. the data structure is described as a great big dictionary, whose value is a pipeline of functors over an initial value. most of these initial values look like `jats('doi')`. the function `jats` here is the scraper wrapping the elife-tools (linked above) function `doi`. The result is then passed through a series of transformations or returned as-is. 
the postprocessing of this data structure after the description has been rendered is similar to how XSLT or XPATH are used with XML: each 'node' in the data structure is visited and the predicate applied. if the predicate is true, the transformation is applied.

I'm not sure where your own preferences lie, but I personally:
* prefer well defined data structures
* prefer lots of little pure functions
* avoid object orientated programming like the plague
* avoid mutable state

I'm more than happy to walk you through any part of my code and explain what it does and why it does it, especially if you find it's purpose or it's design is obscure.

There are a few general purpose utilities in main.py that could be shifted to utils.py but I disagree that it needs refactoring. Shifting it's code into different modules would only serve to spread the logic around and make development/debugging more of a pita.

seanwiseman:
Thank you for the break down of the scraper it was very helpful when re reading the file. With more information it was easier for me to follow.

lsh-0:
my pleasure

PR #343, opened 2017-10-30
Build on c4.4xlarge/c4.8xlarge

giorgiosironi:


giorgiosironi:
Corpus generation goes down from 40-45 minutes measured on `develop` to 23-24. We pay double for the hardware, but we almost halve the time we use it, and since it's billed by the minute this should not cost more.

giorgiosironi:
`c4.8xlarge` takes ~11 minutes for the corpus generation, doing a new run to confirm.

giorgiosironi:
Build times are consistent.

lsh-0:
... (you know, I have an actual Castle of Grayskull in a storage box)

PR #347, opened 2017-11-07
Extract URLs into configuration

Jenniferstrej:
This is a suggestion to have the iiif endpoint configurable via app.cfg. I'm not sure if you still prefer to have the if statements for each environment. Please let me know your opinion. @giorgiosironi @gnott

### DIFF
--- a/elife.cfg
+++ b/elife.cfg
@@ -3,6 +3,8 @@ env: dev
 #upload_path uploads
 # lax--ci needs it's own app.cfg file
 cache_path: /home/elife
+cdn_iiif: https://iiif.elifesciences.org/lax:
+iiif: https://prod--iiif.elifesciences.org/lax:
 
 [lax]
 location: /srv/lax/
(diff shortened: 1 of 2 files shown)


src/conf.py
```diff
@@ -159,18 +159,8 @@ def load(path):
 }
 CDN = 'https://' + CDNS_BY_ENV.get(ENV, DEFAULT_CDN)
```
giorgiosironi: `CDN1`/`CDN` may cause similar problems soon, want to deal with that too?

elife.cfg
```diff
@@ -3,6 +3,8 @@ env: dev
 #upload_path uploads
 # lax--ci needs it's own app.cfg file
 cache_path: /home/elife
+cdn_iiif: https://iiif.elifesciences.org/lax:
```
giorgiosironi: I think the build is failing because of this default, it should be `https://prod--cdn-iiif.elifesciences.org/lax:`. Can you update it?

Jenniferstrej: Yes, sorry for the delay.

PR #361, opened 2017-11-28
Python 3 support

seanwiseman:
The one test failing in both versions since the changes seems to simply return the wrong exception string: 

https://github.com/elifesciences/bot-lax-adaptor/blob/9c093fab702f60f988fb0c0a6c98bf3bf25eb764/src/tests/test_api.py#L200 `AssertionError: u'error-sending-article-json' != 'problem-scraping-xml'`

There are then 5 tests failing because of `request-cache`, there is an [open issue for python 3 support](https://github.com/reclosedev/requests-cache/issues/83) but it seems to be stagnant. We could possibly find an alternative, I did spend a very short amount of time with a fork trying to get to the bottom of it, I stopped on the following: (you will see the full traceback if you run the tests)
```
from dummy_thread import get_ident as _get_ident
ModuleNotFoundError: No module named 'dummy_thread'
``` 
The remaining 2 failures are to do with file paths not existing.


seanwiseman:
@gnott when you can, could you take a look at this?

gnott:
I can take a look, though I use very little of bot-lax-adaptor, just ``main.py`` and ``validate.py`` to get the parser output to be valid. I would suggest adding @lsh-0 to review.

seanwiseman:
Ah OK noted. I have added @lsh-0.

tox.ini
```diff
@@ -0,0 +1,6 @@
+[tox]
+skipsdist = True
+envlist = py27,py35
+[testenv]
+deps = -rrequirements.txt
+commands = coverage run -m unittest discover src/tests
```
giorgiosironi: would expect tox be used in `project_tests.sh` like in https://github.com/elifesciences/elife-tools/blob/develop/project_tests.sh#L4-L7

seanwiseman:
@lsh-0  is there anything else that needs addressing before merging?

PR #377, opened 2017-12-11
[ELPP-3068] lax, only emit an update event on backfill if the article's changed

lsh-0:
ticket: https://elifesciences.atlassian.net/browse/ELPP-3068


lsh-0:
interesting error: 

`sh: 1: cannot create /var/lib/jenkins-libraries-runner/workspace/or_PR-377-G4QPGK6CKRIMIXW4SXAE@tmp/durable-8a650c24/jenkins-result.txt: Directory nonexistent
script returned exit code -1`

giorgiosironi:
Title seemed strange, changed it

giorgiosironi:
`/var/lib/jenkins-libraries-runner/workspace/or_PR-377-G4QPGK6CKRIMIXW4SXAE@tmp` is a workspace, just a folder used by the build, while `durable-8a650c24/jenkins-result.txt` should be a temporary folder and file that capture the output of a shell script. Any disk space problems?

check-hashes.sh
```diff
@@ -0,0 +1,32 @@
...
+function ctrl_c() {
+    exit 1
+}
+
+sumfile="sums.md5"
+
+rm -rf article-json/ article-json-1/ article-json-2
+mkdir -p article-json
+
+python src/generate_article_json.py --num 15
```
giorgiosironi: is this script meant to be used as a test or as an operations script?

src/generate_article_json.py
```diff
@@ -27,11 +27,13 @@ def render(path, json_output_dir):
...
 
-def main(xml_dir, json_output_dir):
+def main(xml_dir, json_output_dir, num=None):
     paths = lmap(lambda fname: join(xml_dir, fname), os.listdir(xml_dir))
     paths = lfilter(lambda path: path.lower().endswith('.xml'), paths)
     paths = sorted(paths, reverse=True)
-    num_processes = 2
+    if num:
+        paths = paths[:num] # only scrape first n articles
+    num_processes = -1
```
giorgiosironi: ah, this should be taken our from this branch and put in master. Being leftover there, it's probably why the bot-lax-adaptor build is taking so long

giorgiosironi:
Just ponting out that this process will maintain a dependency to run on `lax--*--1` (first node if there is a load balancer), like the current backfill, as it relies on the filesystem.

src/generate_article_json.py
lsh-0: > it's probably why the bot-lax-adaptor build is taking so long

yes, we made the change to throttle the rebuilding of the cache file so we didn't overwhelm iiif. I'll update master with this change directly if you haven't already

check-hashes.sh
lsh-0: unless we discover some clever way of determining that article-json output should *not* have changed between revisions, then it's going to require two scrapes to complete. if time were not an issue, I'd scrape everything twice and compare checksums using something like this script, but for now this is just helping me test a subset of articles are identical when scraped twice (they're currently not).

lsh-0: > unless we discover some clever way of determining that article-json output should not have changed between revisions, ...

we could say, if any of the following has changed, run the hash checker:
- requirements.txt
- app.cfg
- src/main.py
- src/conf.py
- src/iiif.py
- src/glencoe.py
- src/cache_requests.py
- src/utils.py

and possibly these as well:
- src/adaptor.py

giorgiosironi: So it's for *testing*, and using a subset makes sense although the subset could be selected to try to provide a good coverage: https://github.com/elifesciences/bot-lax-adaptor/blob/develop/guinea-pigs.sh

giorgiosironi: As in operations usage, I would err on the side of always running the hash checker during a backfill. 4000 articles as input, 4000 articles need to be converted to JSON and validated, but if only 100 actually change, only 100 requests to Lax being made rather than 4000 (and only 100 events are sent).

giorgiosironi: (this particular script is for testing; other could use the same mechanism in operations)

lsh-0:
I think before I merge this I should integrate that hash checking to the project tests. I foresee data changes being unintentionally introduced that are not ordered and will cause lax to send events...

giorgiosironi:
This is still `[wip]` so I'll refrain from going through it, but I agree that the suite should use the hash checking to check for regressions on ordering of fields. Idea:
1. take the guinea pigs from https://github.com/elifesciences/bot-lax-adaptor/blob/develop/guinea-pigs.sh
2. generate them and their hashes
3. do 2 again
4. check the hashes are identical

A different concern: is md5 file hashing the right level of abstraction? JSON dictionaries do not have order (like Python dictionaries), nor whitespace should count; so there may be another algorithm that takes this into account.

giorgiosironi:
e.g. `jq --compact-output --sort-keys "." < elife-12345-v1.xml.json | md5sum` needs jq>=1.4 (we use 1.3), but makes it more canonical

lsh-0:
md5 is useful for doing many files all at once. I just need bot-lax to be doing repeatable scrapes more than anything else, lax stores the article section of the bot-lax scrape as a fragment and then merges with any other fragments. The result is hashed and stored.

We're using OrderedDicts in bot-lax to emit article-json that has a general sort of structure and rely on ordered serialisation/deserialisation into json for that order to be preserved.

I think the hashing for bot-lax is fine as it is, but lax has room for this improvement: https://github.com/elifesciences/lax/blob/5deae7a825f40386d2c9b34d355b96f782ca1678/src/publisher/fragment_logic.py#L190-L192

src/utils.py
```diff
@@ -288,3 +281,13 @@ def ymdhms(dt):
     if dt:
         dt = todt(dt) # convert to utc, etc
         return rfc3339(dt, utc=True)
+
+def sortdict(d):
+    "imposes alphabetical ordering on a dictionary. returns an OrderedDict"
+    if isinstance(d, list):
```
lsh-0: I just had a thought @gnott , that if a `OrderedDict` was encountered, then I could consider the results already ordered and avoid the forced ordering, preserving the ordering efforts you're doing in elife-tools, but it looks like some of the `OrderedDict` results contain unordered nested items. Looks mostly like `additionalFiles` and `figure` elements.

lsh-0:
ready to be reviewed

src/utils.py
lsh-0: sorry, the context is that it's this function sorting everything that is returned from `elife-tools`. I was thinking if it were asked to order an ordereddict I could reasonably assume it was already ordered and return it as-is, but elife-tools is returning OrderedDicts whose contents contain unordered items. Sorry to be confusing.

gnott: Is there a test for ``sortdict()`` that has an ``OrderedDict()`` as the input, and does it sort it or let it pass through in its default order? I guess that's the precise question I was wondering about.

And thanks for clarifying this would sort everything returned from the parser functions.

lsh-0: there are tests: https://github.com/elifesciences/bot-lax-adaptor/pull/377/files#diff-3801658a7002f05f853a58bfb363ab79R94

but not for that specific case. OrderedDicts are sorted. The logic is very simple.

gnott: ``OrderedDict`` sort order would be in the order items are added, and not alphabetical (I believe, I read that again yesterday during some refactoring). If that is sufficient, which it would be for making test assertions, is all cool!  :+1:

PR #414, opened 2018-02-12
PI-626 Allow logging from utils

giorgiosironi:
There is a mutual dependency between the `conf` and `utils` modules, currently with the first including the second. Since the root logger is set up in the first,  can't really log anything as the `LOG` object is setup too soon and doesn't write the log calls.

This PR extracts a `log` module that can be included by in  to break the mutual dependency. As a consequence, some functions have to be moved out of `utils` so that dependencies become like:
- `conf` -> `log`
- `conf` -> `utils`
- `log` -> `dateutils` (new module)
- `utils` -> `dateutils`

### DIFF
--- a/src/dateutils.py
+++ b/src/dateutils.py
@@ -1,0 +1,39 @@
+from datetime import datetime
+from dateutil import parser
+import json
+import pytz
+from rfc3339 import rfc3339
+
+def todt(val):
+    "turn almost any formatted datetime string into a UTC datetime object"
+    if val is None:
+        return None
+    dt = val
+    if not isinstance(dt, datetime):
+        dt = parser.parse(val, fuzzy=False)
+    dt.replace(microsecond=0) # not useful, never been useful, will never be useful.
+
+    if not dt.tzinfo:
+        # no timezone (naive), assume UTC and make it explicit
+        return pytz.utc.localize(dt)
+
+    else:
+        # ensure tz is UTC
+        if dt.tzinfo != pytz.utc:
+            return dt.astimezone(pytz.utc)
+    return dt
+
+def ymdhms(dt):
+    "returns an rfc3339 representation of a datetime object"
+    if dt:
+        dt = todt(dt) # convert to utc, etc
+        return rfc3339(dt, utc=True)
+
+def json_dumps(obj, **kwargs):
+    "drop-in for json.dumps that handles datetime objects."
+    def datetime_handler(obj):
+        if hasattr(obj, 'isoformat'):
+            return ymdhms(obj)
+        else:
+            raise TypeError('Object of type %s with value of %s is not JSON serializable' % (type(obj), repr(obj)))
+    return json.dumps(obj, default=datetime_handler, **kwargs)
--- a/src/log.py
+++ b/src/log.py
@@ -1,0 +1,57 @@
+import logging
+from pythonjsonlogger import jsonlogger
+import dateutils
+
+def setup_root_logger():
+    root_logger = logging.getLogger("")
+
+    # output to stderr
+    _handler = logging.StreamHandler()
+    _handler.setLevel(logging.INFO)
+
+    class FormatterWithEncodedExtras(logging.Formatter):
+        def format(self, record):
+            # exclude all known keys in Record
+            # bundle the remainder into an 'extra' field,
+            # bypassing attempt to make Record read-only
+            _known_keys = [
+                'asctime', 'created', 'filename', 'funcName', 'levelname', 'levelno', 'lineno',
+                'module', 'msecs', 'message', 'name', 'pathname', 'process', 'processName',
+                'relativeCreated', 'thread', 'threadName',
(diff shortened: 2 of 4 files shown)


src/dateutils.py
```diff
@@ -0,0 +1,39 @@
...
+from rfc3339 import rfc3339
+
+def todt(val):
+    "turn almost any formatted datetime string into a UTC datetime object"
+    if val is None:
+        return None
+    dt = val
+    if not isinstance(dt, datetime):
+        dt = parser.parse(val, fuzzy=False)
+    dt.replace(microsecond=0) # not useful, never been useful, will never be useful.
```
seanwiseman: I read this as the statement is not useful and should, therefore, be removed. Or is it referring to the microsecond value being useless?

src/log.py
```diff
@@ -0,0 +1,57 @@
...
+            record.__dict__['extra'] = dateutils.json_dumps(unknown_fields)
+            return super(FormatterWithEncodedExtras, self).format(record)
+
+    _handler.setFormatter(FormatterWithEncodedExtras('%(levelname)s - %(asctime)s - %(message)s -- %(extra)s'))
+
+    root_logger.addHandler(_handler)
+    root_logger.setLevel(logging.DEBUG)
+
+def json_formatter():
+    _supported_keys = [
```
seanwiseman: Remove the unused items? also, the leading `_` in the variable names are redundant in this use case.

giorgiosironi:
Waiting for all the comments before the refactoring phase.

src/dateutils.py
lsh-0: it's referring to the microsecond value being useless

src/log.py
lsh-0: the commented out items serve to show us what _can_ be included vs what _is_ being included.

the underscores denoting private/public scope are something I tend to avoid so I'm not 100% sure this was my code, but I've never applied it consistently. ultimately it doesn't really bother me.

giorgiosironi:
Will clean up a bit before merging. To really solve the problems consistently across projects, we may need an equivalent of https://github.com/elifesciences/logging-sdk-php (now I will find out that it already exists and I don't remember)

seanwiseman:
In your opinions @giorgiosironi @lsh-0 , what are the problems with the current builtin `logging` library for python? as I don't have a problem with it, but have had issues using other peoples attempts and recreating it on other projects.

Just interested.

src/log.py
giorgiosironi: The underscores are there because this code comes from the top level of a module, but can now be removed

lsh-0:
I have problems with the model in general, which go back to 2001 with log4j and probably further.

* log output (formatters) is semi-structured at best. 
* the json logging I've introduced seems crude but gets the job done.
* 'best practice' is that each module gets it's own logger but ...
* the logging module is stateful
* logging out of the box requires elaborate setup. elaborate setup must be performed for each logger unless logging configuration is shared, in which case order of configuration become important.
* most libraries come with their logging silenced, but those libraries descend from the root logger, so if your config involves setting global rules for the root logger, you're affecting those libraries as well. see boto

and often it just doesn't fucking work when you need it too. "Why isn't it working today? It has a formatter and a handler and a name and the destination file exists and I'm following best practice and I've prayed to the gods and still nothing! why does `basicConfig` work and not this? do I need to ..." - just way more effort than necessary.

ideally:

```python
./python
>>> from statelesslogging import log
>>> log.info("pants") # default stderr
{'level': 'info', 'msg': 'pants', 'module': '__main__', 'line': 1, ...}
>>> from statelesslogging import filelog
>>> log = filelog('log.json')
>>> log.info("pants")
>>> json.load(open('log.json'))
{'level': 'info', 'msg': 'pants', 'module': '__main__', 'line': 5, ...}
```

seanwiseman:
Very interesting, thank you @lsh-0 you raise some good points. Definite food for thought.

giorgiosironi:
> the logging module is stateful
> logging out of the box requires elaborate setup. elaborate setup must be performed for each logger unless logging configuration is shared, in which case order of configuration become important.

These are my main concerns, global state (which lead to this bug):
```
import logging
LOG = logging.getLogger(__name__)
LOG.info("something") # doesn't log anything despite me having configured logging in a "bootstrap" somewhere in the app, because that bootstrap is not guaranteed to have been imported
import conf # the bootstrap
LOG.info("something") # still doesn't log anything because LOG was created too soon
```
I work around this by creating a `log` module that anyone that wants to log has to import, and set up logging in that module. Still means some library logging calls occurring too early may be lost. The `logging` module is a standard and it won't go away, but I'd like to push its setup somewhere else so that it will "just work" when required by a project; bonus point, it will become consistent across projects so when debugging there will always be a JSON file with certain keys, etc.

seanwiseman:
Thank you @giorgiosironi. Both these comments have lit a small fire beneath me regarding the current `logging` strategies and methods.

PR #422, opened 2018-03-01
Add "of" to slugify stopwords for the subject id.

gnott:


giorgiosironi:
Interesting to see where this comes from. Makes the case for an even faster pipeline to bring these changes in production.

PR #426, opened 2018-03-06
removed reference to elife-hosted fork of connexion

lsh-0:
doesn't look like it's being used and I want to delete https://github.com/elifesciences/connexion


lsh-0:
yeah, it's used to generate and validate the api bot-lax is using to allow uploads of xml with custom key:vals and send them to lax for validation. We had some problems with it for a time, some kind of incompatibility with another library, but I haven't seen it being a nuisance in a while.

PR #433, opened 2018-03-19
Added xml_uri generation

seanwiseman:
A task from https://github.com/elifesciences/api-raml/pull/215.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -172,8 +172,6 @@ def pdf_uri(triple):
 def xml_uri(triple):
     """predict an article's xml url."""
     content_type, msid, version = triple
-    if content_type and any(lmap(lambda type: type in ['Correction', 'Retraction'], content_type)):
-        return EXCLUDE_ME
     filename = "elife-%s-v%s.xml" % (utils.pad_msid(msid), version) # ll: elife-09560-v1.xml
     return cdnlink(msid, filename)
 
--- a/src/main.py
+++ b/src/main.py
@@ -169,9 +169,9 @@ def pdf_uri(triple):
     filename = "elife-%s-v%s.pdf" % (utils.pad_msid(msid), version) # ll: elife-09560-v1.pdf
     return cdnlink(msid, filename)
 
-def xml_uri(triple):
+def xml_uri(params):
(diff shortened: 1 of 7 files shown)


src/main.py
```diff
@@ -169,6 +169,14 @@ def pdf_uri(triple):
     filename = "elife-%s-v%s.pdf" % (utils.pad_msid(msid), version) # ll: elife-09560-v1.pdf
     return cdnlink(msid, filename)
 
+def xml_uri(triple):
+    """predict an article's xml url."""
+    content_type, msid, version = triple
+    if content_type and any(lmap(lambda type: type in ['Correction', 'Retraction'], content_type)):
```
giorgiosironi: this is true for PDF (where this comes from) but I think the XML is always there for any article. I'll check.

seanwiseman: Ok thank you.

src/main.py
```diff
@@ -169,6 +169,14 @@ def pdf_uri(triple):
...
+        return EXCLUDE_ME
+    filename = "elife-%s-v%s.xml" % (utils.pad_msid(msid), version) # ll: elife-09560-v1.xml
```
giorgiosironi: it is correct to pad this:
```
[14:34:25][giorgio@Newton:~/code/builder]$ aws s3 ls prod-elife-published/articles/09560/ | grep xml
2017-03-29 06:47:41     249611 elife-09560-v1-download.xml
2017-03-29 06:47:41     249611 elife-09560-v1.xml
```
:+1:

src/main.py
```diff
@@ -169,6 +169,12 @@ def pdf_uri(triple):
     filename = "elife-%s-v%s.pdf" % (utils.pad_msid(msid), version) # ll: elife-09560-v1.pdf
     return cdnlink(msid, filename)
 
+def xml_uri(triple):
+    """predict an article's xml url."""
+    content_type, msid, version = triple
```
giorgiosironi: content-type is not needed anymore here

seanwiseman: You are right. I have now removed it. I felt pressured to keep it due to the name `triple` 😱.

PR #463, opened 2018-06-19
Use canonical IIIF hostname

giorgiosironi:
Just changed this in the `prod` configuration, so reflecting it in testing.

### DIFF
--- a/src/iiif.py
+++ b/src/iiif.py
@@ -5,7 +5,7 @@ import requests_cache
 from cache_requests import install_cache_requests
 import conf, utils
 
-LOG = logging.getLogger(__name__)
+LOG = conf.multiprocess_log('iiif.log', __name__)
 
 if conf.REQUESTS_CACHING:
     install_cache_requests()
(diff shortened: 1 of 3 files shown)


src/iiif.py
```diff
@@ -5,7 +5,7 @@
 from cache_requests import install_cache_requests
 import conf, utils
 
-LOG = logging.getLogger(__name__)
+LOG = conf.multiprocess_log('iiif.log', __name__)
```
giorgiosironi: at least locally, this logger seemed to be configured not to write to a file nor stderr

src/iiif.py
```diff
@@ -49,7 +49,9 @@ def iiif_info(msid, filename):
         'iiif_info_url': iiif_info_url(msid, filename)
     }
     try:
-        resp = utils.requests_get(iiif_info_url(msid, filename))
+        url = iiif_info_url(msid, filename)
+        LOG.info("Loading IIIF info URL: %s", url)
```
giorgiosironi: this should not change, and should not be the CDN since it's loaded from inside the VPC

PR #466, opened 2018-06-28
truncates length of message to be sent bia SQS to first 256k characters.

lsh-0:
ticket: https://elifesciences.atlassian.net/browse/ELPP-3740

* sqs message is *always* deleted, preventing the infinite loop case
* further validation of response to ensure size is something SQS can manage

### DIFF
--- a/src/sqs_adaptor.py
+++ b/src/sqs_adaptor.py
@@ -24,8 +24,11 @@ def poll(queue_obj):
                 WaitTimeSeconds=20 # maximum setting for long polling
             )
         message = messages[0]
-        yield message.body
-        message.delete()
+        try:
+            yield message.body
+        finally:
+            # message is always deleted, even if this leads to no response to sent messages
+            message.delete()
 
 class IncomingQueue(object):
     def __init__(self, queue_name, flag=None):
@@ -57,10 +60,6 @@ class OutgoingQueue(object):
 
     def write(self, string):
         "called when given a validated response"
-        # totally naive. unicode check probably needed
-        if len(string) >= 256000:
-            LOG.error("message to be sent is very large and will be truncated to 256KB")
-        string = string[:256000]
         self.queue.send_message(MessageBody=string)
 
     def error(self, string):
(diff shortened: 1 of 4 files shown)


src/sqs_adaptor.py
```diff
@@ -57,6 +57,10 @@ def __init__(self, queue_name):
 
     def write(self, string):
         "called when given a validated response"
+        # totally naive. unicode check probably needed
+        if len(string) >= 256000:
```
thewilkybarkid: Reading https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-limits.html#w117aac20b9b5 it's actually KiB, so 262144 (though I suspect the missing 6144 bytes won't be that useful).

src/sqs_adaptor.py
```diff
@@ -57,6 +57,10 @@ def __init__(self, queue_name):
...
+            LOG.error("message to be sent is very large and will be truncated to 256KB")
+        string = string[:256000]
```
giorgiosironi: given this is usually a JSON document, truncating it like this would make it unparseable so not very useful on the other side? We could try to go through the fields and truncate those instead, but I suspect it's easier to go to the source and truncate the validation error message when it is produced, even if it introduces some knowledge of SQS into that part

lsh-0:
> it's actually KiB, so 262144

nice catch! I re-read the docs and they definitely conflate KB with KiB

src/sqs_adaptor.py
```diff
@@ -24,8 +24,11 @@ def poll(queue_obj):
                 WaitTimeSeconds=20 # maximum setting for long polling
             )
         message = messages[0]
-        yield message.body
-        message.delete()
+        try:
+            yield message.body
+        finally:
+            # message is always deleted, even if this leads to no response to sent messages
```
giorgiosironi: great :+1:

giorgiosironi:
Nice to have: if another default error message could be sent so that it appears on the dashboard. Not sure if it's a good idea when things are this broken.

PR #467, opened 2018-07-02
api, remove path of uploaded file

lsh-0:
api, fixed case where path to xml was being included in final path for uploaded xml


lsh-0:
fyi @giorgiosironi and @gnott - a bug that Exeter encountered. if it passes I'll merge.

lsh-0:
huh. haven't seen scrape errors in a long time. I forked from latest in develop so the problem is probably there as well.

lsh-0:
I tested on another branch if an update to api-raml was the cause of the failure (it was), then merged that into develop and then develop into this.

gnott:
I hope a bad path similar to what you included in the test case will be caught much earlier because it probably had more implications. It would be bad data for sure, but it is probably not the responsibility for bot-lax-adaptor to report the error anyway.

lsh-0:
I'm not 100% sure how Exeter were making the request, but that path was a path on their filesystem, presumably the file to be uploaded, and the bot-lax api had been tested only on local files, or on tweaked POST data to make it look like the test fixture was local. It also looks like I knew about the problem ages ago and did nothing about it, so, shame on me.

PR #470, opened 2018-07-04
[WIP] Fix IIIF identifiers

thewilkybarkid:


thewilkybarkid:
Loris appears to be overloaded. Didn't realise this would trigger a full cache-busting rebuild. 😒

giorgiosironi:
> a full cache-busting rebuild

I think a full rebuild of all the articles is part of the test suite for `bot-lax-adaptor`, but it manages to cache locally the `info.json` of most images in normal cases; the cache key is the URL however

giorgiosironi:
Build seems to have failed on a third of the articles due to the IIIF overload, if a re-run doesn't manage to work with the primed cache we'll introduce `backoff`.

giorgiosironi:
Cache fully populated now

thewilkybarkid:
Should be mergeable.

PR #473, opened 2018-07-09
Feature: api relax fname requirement

lsh-0:
ticket: https://github.com/elifesciences/jira-import/issues/4307

It is inconvenient for third parties to use the elife filename format (elife-{msid}-v{version}.xml) when sending article xml to the bot-lax api.

This change adds the ability to supply the msid and version as query parameters to the `/xml` endpoint. The versions need not match, but the msid should.

### DIFF
--- a/schema/api.yaml
+++ b/schema/api.yaml
@@ -78,7 +78,17 @@ paths:
           uniqueItems: true
           items:
             type: string
-            
+        - name: id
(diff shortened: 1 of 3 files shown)


lsh-0:
yes, still required. this is a change to the api only. The api validates generated article-json only, it doesn't change anything in lax.

PR #476, opened 2018-07-10
Update IIIF prefix

thewilkybarkid:


gnott:
I'm seeing both URI work, so do we not need a full Lax backfill right away?

thewilkybarkid:
Yeah, there's no rush.

PR #486, opened 2018-08-13
added check for writable requests-cache dir

lsh-0:
I'm switching from my laptop to a desktop and found a nuisance


lsh-0:
running the tests with the default (`/home/elife`) had the tests failing to start cryptically when run locally.

I'll give the build a kick now that Alfred is feeling better

lsh-0:
kicking the build, it's been (almost!) a full year

lsh-0:
my fix is disabling that key in the default `elife.cfg` file so that `conf.py` will use it's own (better) default.

lsh-0:
total derp. that value isn't a directory at all, it's a file and `requests_cache` appends the suffix `.sqlite3` so the (default) path is `./cache/requests_cache.sqlite3`

closing PR

PR #498, opened 2018-10-05
cdn.py now clears non-200 cached responses

lsh-0:
code was already there, looks like we've used it before. I just gave it a brush off and stuck it in the error handling

### DIFF
--- a/src/cdn.py
+++ b/src/cdn.py
@@ -11,6 +11,8 @@ LOG = logging.getLogger(__name__)
 if conf.REQUESTS_CACHING:
     install_cache_requests()
 
+def clear_cache(url):
+    conf.REQUESTS_CACHING and requests_cache.core.get_cache().delete_url(url)
 
 def url_exists(url, msid=None):
     context = {'msid': msid, 'url': url}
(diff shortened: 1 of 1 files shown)


gnott:
Does the CDN only return `200` or is a `203` possible? Just wondering.

src/cdn.py
```diff
@@ -25,14 +27,16 @@ def url_exists(url, msid=None):
...
-    elif resp.status_code == 404:
+
+    # non-200 response
+    # a request outside of the regular workflow may have been made too early
+    # a subsequent request would return the 'not found' response
+    # https://github.com/elifesciences/issues/issues/4458
+    clear_cache(url)
+
+    if resp.status_code == 404:
         LOG.debug("CDN url not found", extra=context)
```
gnott: I may not understand completely, trying to step through it and from my limited knowledge of `requests_cache` - in the above `resp = requests.head(url)`, would that end up being cached automatically? Maybe `clear_cache(url)` should go here if the status code is 404 then clear the cache of it immediately?

lsh-0:
a 203 may be possible but it wouldn't have worked with the previous code which was checking for 200 and 404 explicitly and raising an error on anything else

src/cdn.py
gnott: Oh, duh, nevermind - yours has the same effect, clearing any non-200 :+1:

lsh-0: I'm making the distinction between successful and unsuccessful responses, so 200 == success, cache and forget. Anything else is unsuccessful, which is unusual, so wipe state so there is no uncertainty about subsequent requests.

lsh-0: no worries :see_no_evil:

lsh-0:
>  however there is also a CDN layer (for some requests) 

this change isn't affecting all requests made, just those being made through `cdn.py`. Is this what you meant? Or are we talking about external caching in the CDN? 

I suppose figures and perhaps even glencoe requests would suffer the same problem we have here. One fix at a time

lsh-0:
I'll bump the version in lax. There is an unrelated dependency vulnerability for lax as well, so I'll bump all of it's requirements at once

giorgiosironi:
> this change isn't affecting all requests made, just those being made through cdn.py

Anything that goes to `cdn.elifesciences.org`, as the CDN has its own cache.

PR #500, opened 2018-10-19
feature: better glencoe error

lsh-0:
improved glencoe error message on empty response.
only successful glencoe responses are cached

### DIFF
--- a/src/glencoe.py
+++ b/src/glencoe.py
@@ -65,15 +65,20 @@ def clear_cache(msid):
     requests_cache.core.get_cache().delete_url(glencoe_url(msid))
 
 def metadata(msid):
-    resp = utils.requests_get(glencoe_url(msid))
-    context = {'msid': msid, 'status-code': resp.status_code}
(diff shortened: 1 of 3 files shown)


src/glencoe.py
```diff
@@ -65,15 +65,20 @@ def clear_cache(msid):
     requests_cache.core.get_cache().delete_url(glencoe_url(msid))
 
 def metadata(msid):
-    resp = utils.requests_get(glencoe_url(msid))
-    context = {'msid': msid, 'status-code': resp.status_code}
+    # 2018-10-19: it's now possible for glencoe to be queried about an article before media
+    # has been deposited by elife-bot. only successful responses will be cached
+    url = glencoe_url(msid)
+    resp = utils.requests_get(url)
```
giorgiosironi: need a `LOG.info` of `resp` here, so that if there is a case that has fallen into the tracks like 403 or a 200 with a strange body, we'll collect a dump of it

PR #505, opened 2018-11-12
Updated elife-article-xml to: 2018-11-09 16:55:29 +0000

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-bot-lax-adaptor-update-elife-article-xml/93/display/redirect which resulted in this pull request.


giorgiosironi:
One article from the corpus is now failing:
```
INFO - 2018-11-12 06:46:25,999 - elife-12572-v3.xml -> elife-12572-v3.xml.json => failed (glencoe doesn't know 'video1', only 'media1, media2, media3, media4')
```
@lsh-0 @gnott is the data from Glencoe incorrect?

lsh-0:
related to this: https://github.com/elifesciences/issues/issues/4493

lsh-0:
v2 is using 'media1' etc, v3 is using 'video1' etc.

I suggest the XML for v3 be fixed up and a silent correction made

gnott:
https://github.com/elifesciences/issues/issues/4493#issuecomment-430421117

> If you were to, for example, load the v2 of this article again via a silent correction, it will fail because its videos are no longer named what they were in 2016.

> I suggest you might want to change the v2 to have the more recent video naming in the XML ids and the file names, and load that as a silent correction, to avoid some future content discord.

gnott:
There could be some mismatch, they may have all been changed back to `media1` in the v3 according to the issues ticket, not too sure yet.

gnott:
You can probably ignore my comments, I think you've resolved it (still catching up on issues I've missed).

PR #542, opened 2019-04-23
Updated elife-tools to: requirements.txt pinned to 63fa9cfc9a8eb233a05b37279def7e15531816e9

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies-bot-lax-adaptor-update-elife-tools/101/display/redirect which resulted in this pull request.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ boto3==1.8.7
 connexion==1.5.2
 coverage==4.1
 et3==1.5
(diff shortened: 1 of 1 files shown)


gnott:
This PR is the result of merging https://github.com/elifesciences/elife-tools/pull/305, which was supposed to be an improvement for handling comment tag parsing. 

The failure of the test case here shows a slight regression, where previously content returned `None` is it returning a blank string using the newest `elifetools`.

I'm going to merge in a fix to `elifetools` for it to return `None` as it did before, to avoid any breaking changes due to the comment tag logic that was changed.

PR #567, opened 2019-08-20
Feature: disable glencoe caching

lsh-0:
* adds ability to disable glencoe requests_cache response caching
* git-ignores hidden pytest dir
* ~disables `cache_path` setting with (useless) value of `/home/elife`~

- [x] blocked by https://github.com/elifesciences/lax-formula/pull/37

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,5 @@ elife-tools/
 src/elifetools
 /*.sqlite3
 run-*/
+.pytest_cache/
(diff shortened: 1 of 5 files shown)


lsh-0:
I think `/home/elife/` is actually the home of the cache and doesn't have an env override ...

lsh-0:
I think all the alerts just now was this PR + corpus generation thrashing things

PR #576, opened 2019-09-17
Feature: added ability to retry a http get on 404 response

lsh-0:
This allows the request to Glencoe be made twice more if the original response is a 404 with the standard exponential backoff, starting at 1 second, then 2 then 4.

It only affects requests made by glencoe.py

### DIFF
--- a/src/glencoe.py
+++ b/src/glencoe.py
@@ -70,10 +70,10 @@ def metadata(msid):
     url = glencoe_url(msid)
 
     if conf.REQUESTS_CACHING and conf.GLENCOE_REQUESTS_CACHING:
-        resp = utils.requests_get(url)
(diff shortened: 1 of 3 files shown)


lsh-0:
>  Looks like they have a consistency problems where different servers give different results

That's my gut feeling too, but without knowing specifics this is the best for now. If we keep seeing it I'll contact them and ask them to take a look

lsh-0:
actually, forgot to add a log statement so we can keep track of when this happens

gnott:
I'm cleaning up some stale branches, @lsh-0 is this PR still under consideration?

lsh-0:
> That's my gut feeling too, but without knowing specifics this is the best for now. If we keep seeing it I'll contact them and ask them to take a look

I don't recall any further issues about talking to glencoe.

PR #596, opened 2020-01-10
removes reference to python2.7 in canary.sh

lsh-0:


lsh-0:
canary script is still working nicely actually.

I just did a run through and the latest dependencies work quite nicely. separate PR though

PR #603, opened 2020-02-11
Updated elife-tools to: requirements.txt pinned to ccc610adf7d658642399d49d1f5906057487606b

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies-bot-lax-adaptor-update-elife-tools/109/display/redirect which resulted in this pull request.


gnott:
I think this could be a result of `py36` changes and not `elife-tools` _per se_ @lsh-0 ?

lsh-0:
actually it's not :) 

    [2020-02-11T03:41:38.755Z]     from werkzeug import cached_property
    [2020-02-11T03:41:38.755Z] E   ImportError: cannot import name 'cached_property'

this is a relatively recent error, started happening about Monday iirc. Flask-testing and Werkzeug became incompatible and the version of werkzeug needs to be pinned until they fix things.

I'll give this build a kick because bot-lax already has that fix in it now

PR #627, opened 2020-06-02
Structured abstract demonstration

gnott:
I tried to assemble a very simple example of a structured abstract, with a simplified article fixture, a draft `elife-tools` release, and development branch of the API RAML.

It doesn't solve the issue of downgrading a structured abstract to an earlier schema version, but hopefully it is something to begin with, or this branch / PR can be deleted and started fresh with something borrowed from here.

@lsh-0 if you have ideas or would like to collaborate on developing the branch?


lsh-0:
I'm being kept pretty busy already and don't have any structured abstract work assigned to me. I'm happy to review and answer questions and at some point there will be some lax integration work required but until then I don't have any capacity to take on extra work.

gnott:
Sounds fine to me, thanks for the details Luke. After have a day to think about this side of things, there's probably not much more required here. The bare minimum will be to update the sha of `elife-tools` and `api-raml`, and the additional test case is a little extra if we want it.

I think the main effort will happen in Lax and all the other projects. This branch can at least demonstrate the article JSON produced is valid against a schema, and that's about it.

lsh-0:
I think the api-raml pinned revision in this branch is old or wrong now, I get an error when trying to use it.

can this PR be closed?

gnott:
As a test, I'm switching back some values now the `api-raml` change is in its master branch. I think the tests on this PR have always been green, and I'd expect that again (tests are currently waiting for resources currently blocked by other tests).

Beyond the simple test case I added which has a structured abstract, there's nothing new in this PR, except the bump in the schema article versions it checks against. So, once my curiosity of whether the tests remain to be passing, I'm happy to close this, because it adds little or no value to the base code.

gnott:
Yes, still green, that's good to see. But, we don't need this PR, so now closing it.

PR #647, opened 2020-07-06
Updated api-raml to: 2020-07-06 03:03:02 +0000

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-bot-lax-adaptor-update-api-raml/250/display/redirect which resulted in this pull request.


gnott:
@lsh-0 PRs in this project were not green, I briefly looked. The first runs seemed to have a problem with

`error: invalid command 'bdist_wheel'`

which I guessed was probably temporary, so I re-ran the tests on this PR. They made it beyond that point I think, but may have gotten stuck at the `pylint` stage.

I also noticed Alfred was not able to connect to port 25 for SMTP, which might be normal for this project, or maybe it indicates a reason for the failure, due to instance config or networking?

lsh-0:
>  but may have gotten stuck at the pylint stage.

I fixed an issue with lax yesterday with a similar problem. The problem was that a dependency saw a major-version upgrade and broke linting. The linting has some unhelpful /dev/null redirect in it so it's not clear how it's failing.

The bdist wheel thing is more of a warning than an error, and I'm not sure about any SMTP server on Alfred, but I guess it has to be sending emails somehow.

lsh-0:
I'll fix this up now

gnott:
Thanks @lsh-0, it is green, I will merge it I think.

PR #652, opened 2020-07-21
Data watch display channel.

FAtherden-eLife:
Data watch display channel in XML mapped to `tools-resources`.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -86,7 +86,7 @@ DISPLAY_CHANNEL_TYPES = {
     "Research article": "research-article",
     "Short report": "short-report",
     "Short Report": "short-report",
-    "Tools and Resources": "tools-resources",
+    "Data watch": "tools-resources",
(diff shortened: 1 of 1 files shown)


src/main.py
```diff
@@ -86,7 +86,7 @@ def fn(soup):
     "Research article": "research-article",
     "Short report": "short-report",
     "Short Report": "short-report",
-    "Tools and Resources": "tools-resources",
```
gnott, line 89: You could leave in this line if you want, since it is possible to have multiple phrases mapping to the same display channel.

FAtherden-eLife: Thanks - Re-added in the latest commit.

gnott:
Looks good, thanks @FAtherden-eLife. Attempts to make all the tests pass are not working, and I don't expect they will anyway for `ijm` branch. Merging!

PR #672, opened 2020-10-19
Use latest article schema model when validating.

gnott:
Re issue https://github.com/elifesciences/issues/issues/6222

This may solve an issue with pipeline test failures on structured abstracts.

We may not have any real structured abstracts to test this against, but if this is green and it is a suitable change, we can test against real structured abstracts after merging this as a fix.

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -91,8 +91,8 @@ def load(path):
     elif path.endswith('.yaml'):
         return yaml.load(open(path, 'r'))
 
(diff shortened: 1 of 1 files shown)


gnott:
A similar config change I suggested in PR https://github.com/elifesciences/bot-lax-adaptor/pull/627 which we closed.

lsh-0 (approved):

gnott:
I will give this a go to see if it fixes the other PR

PR #679, opened 2020-11-20
adds support for reading XML from S3 and an example usage script

lsh-0:
* adds support for reading article XML from an S3 bucket
* adds a script with example usage (dry-run/validate-only)

### DIFF
--- a/reload-article-from-s3.sh
+++ b/reload-article-from-s3.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -eu
+
+# https://s3-external-1.amazonaws.com/bucket/path/to/elife-59976-v2.xml"
+target=$1
+
+# `--type s3` how to resolve `target`. can also handle the filesystem and AWS SQS
+# `--action ingest+publish` does an ingest and the a publish in one action
+# `--force` ignore publication status
+# `--validate-only` do a 'dry run' and do not commit the transaction
+python src/adaptor.py \
+    --type s3 \
+    --target "$target" \
+    --action ingest+publish \
+    --force \
+    --validate-only
--- a/src/adaptor.py
+++ b/src/adaptor.py
@@ -331,6 +331,15 @@ def read_from_fs(path, **kwargs):
     outgoing = fs_adaptor.OutgoingQueue()
     return incoming, outgoing
 
+def read_from_s3(path, kwargs):
+    "generates a message for a file on s3 like SQS, but without using SQS"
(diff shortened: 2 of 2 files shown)


lsh-0:
this is intended for debugging purposes, specifically the issue we're seeing now in 6209

src/adaptor.py
```diff
@@ -331,6 +331,15 @@ def read_from_fs(path, **kwargs):
     outgoing = fs_adaptor.OutgoingQueue()
     return incoming, outgoing
 
+def read_from_s3(path, kwargs):
```
gnott, line 334: I'm not entirely sure, but would `kwargs` possibly conventionally be `**kwargs` in this situation?

gnott: Guessing maybe since you pass it later, you do not want to "unpack" the values early?

reload-article-from-s3.sh
```diff
@@ -0,0 +1,16 @@
...
+target=$1
+
+# `--type s3` how to resolve `target`. can also handle the filesystem and AWS SQS
+# `--action ingest+publish` does an ingest and the a publish in one action
+# `--force` ignore publication status
+# `--validate-only` do a 'dry run' and do not commit the transaction
+python src/adaptor.py \
+    --type s3 \
+    --target "$target" \
+    --action ingest+publish \
```
gnott, line 14: A comment / question, I seem to remember on the `elife-bot` side of things we stopped sending `ingest+publish` in any of the ingest or publish messages. I'm not sure why, but it might have had something to do with when we added `force: true` then it didn't make any difference? I cannot remember, and I might have this wrong. I'm wondering if this would test the particular situation we observed recently. Of course, since this script is an example, it doesn't matter what the `action` value is here.

gnott (approved):
This was an opportunity for me to look at how things connect together, and your changes look ok to me as far as I could tell. Good luck with the testing!

reload-article-from-s3.sh
lsh-0: The script is an example but I'll be tweaking it until I can pull the new xml down from s3 and pushed it into lax. I might even find the source of the bug in the process but if I don't at least it will unblock the article.

lsh-0: I don't know why `ingest+publish` isn't being used, probably because it's unnecessary day to day.

src/adaptor.py
lsh-0: I see what you're saying, it could be either-or. I prefer to avoid the `*` and `**` operators these days.

PR #708, opened 2021-02-25
Updated api-raml to: 2021-02-25 17:35:35 +0000

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-bot-lax-adaptor-update-api-raml/291/display/redirect which resulted in this pull request.


gnott:
My thought is this will be fine to merge as is because on its own it is backwards compatible while certain journal references are still converted to type `unknown` by the XML parser, and also validation is not set to use the `v5` of VoR article definition anywhere. Subsequent PRs will be testing the schema changes for real.

PR #713, opened 2021-03-05
backfill, adds script for backfilling many articles.

lsh-0:
backfill.sh and backfill-single.sh weren't suitable and neither was calling ./src/adhoc-backfill.sh with hundreds/thousands of paths.

using `backfill-many.sh` you add a list of manuscript IDs to the file `backfill.txt` and then run `./backfill-many.sh` with no arguments.

there will be a complete backfill coming up soon with the preferred name change. I'll need to figure out a way to do that that doesn't bring lax to it's knees.

### DIFF
--- a/src/backfill-many.py
+++ b/src/backfill-many.py
@@ -8,7 +8,7 @@ import adhoc_backfill
 
 def main(xml_repo_dir, backfill_file, dry_run):
     assert os.path.exists(xml_repo_dir), "path to XML repository doesn't exist: %s" % xml_repo_dir
-    
+
     path_list = []
     for globbed_path in open("backfill.txt", 'r').read().splitlines():
         globbed_path = os.path.join(xml_repo_dir, globbed_path)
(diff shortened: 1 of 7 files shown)


src/backfill-many.py
```diff
@@ -0,0 +1,52 @@
+# backills many articles using a simple text file of paths to XML.
+# paths may include globs.
+# all paths are verified to exist before execution.
+# see `/backfill-many.sh`
+
+import glob, os, sys
+import adhoc_backfill
+
+def main(xml_repo_dir, backfill_file, dry_run):
```
gnott, line 9: `backfill_file` argument is perhaps not used yet in this function, maybe goes in place of "backfill.txt" on line 13?

lsh-0: good catch

lsh-0:
merging this so I can see what the outcome is of backfilling those articles with broken IIIF files.

I assumed they failed to backfill properly but I could be mistaken. It should be known before we attempt the complete backfill for the unrelated api-raml change

PR #715, opened 2021-03-08
Updated api-raml to: 2021-03-04 08:16:22 +0000

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-bot-lax-adaptor-update-api-raml/295/display/redirect which resulted in this pull request.


gnott:
This will merge in using `api-raml` version `2.0.0.`

@lsh-0, is there anything we should do before merging this PR please?

lsh-0 (approved):
> @lsh-0, is there anything we should do before merging this PR please?

I don't think so. `2.0.0` removes older versions of article content types and everything that relied on article data should have been updated to use VORv5 and POAv3 by now.

lsh-0:
If a project hasn't been updated and is not pinned, then it's just begging for trouble. I don't know of any project like that.

PR #717, opened 2021-03-10
chore: update project maintainers

NuclearRedeye:


lsh-0 (approved):
just an fyi, jenkins doesn't support aliases yet, it's in this sprint though

PR #747, opened 2021-06-21
Updated api-raml to: 2021-06-21 03:49:55 +0000

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies/job/dependencies-bot-lax-adaptor-update-api-raml/313/display/redirect which resulted in this pull request.


gnott:
I think this is ok to merge, just in case @lsh-0 if you would please confirm?

lsh-0:
is this the raml with the new preprint work? if so, it should be ok. I'll be working on it's implementation today probably

gnott:
According to the sha this is the latest, including preprint, I will merge it, thanks!

PR #748, opened 2021-06-23
Updated elife-tools to: requirements.txt pinned to 31509bfc9a8b84656ed6a1e8433894c56eff0b83

elife-alfred-user:
I have run https://alfred.elifesciences.org/job/dependencies-bot-lax-adaptor-update-elife-tools/142/display/redirect which resulted in this pull request.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ boto3==1.8.7
 connexion==1.5.2
 coverage==4.1
 et3==1.5.1
-git+https://github.com/elifesciences/elife-tools.git@4301fd8a63ab95819f1a2936a20f0811c2fabb75#egg=elifetools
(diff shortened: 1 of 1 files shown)


gnott:
A note, I think this is just noise, because it is functionally equivalent to the previous sha commit which included the pub history parsing changes. This PR is a result of me noticing the `master` branch of `elife-tools` was one commit ahead of `develop`, and bringing the `develop` branch up-to-date has resulted in this.

We should be able to get rid of this noise after transitioning entirely over to versioned releases used in pipelines, and to stop relying on commits to the `develop` branch of `elife-tools`. @lsh-0 an FYI, which I'm sure you're already aware.

It should also mean this PR, if green, is safe to merge.

lsh-0:
all good, thanks for the heads-up @gnott

PR #749, opened 2021-06-25
preprints, adds parsing of pub_history to extract the first preprint

lsh-0:


lsh-0:
hey @gnott , if you could let me know if you think I've missed anything I'd be grateful.

The changes include some fixing some deprecation warnings and removing some old old commented out code.

src/main.py
```diff
@@ -262,6 +253,22 @@ def body(soup):
 def appendices(soup):
     return jats('appendices_json', base_url(jats('publisher_id')(soup)))(soup)
 
+def preprint_events(struct):
+    "returns a list of 'preprint' type events from article's pub-history or `None`"
+    if not struct or not isinstance(struct, list) or len(struct) == 0:
+        return
+    return [event for event in struct if event.get('type') == 'preprint'] or None
```
gnott, line 260: I think possibly I named this `event_type`, i.e. `if event.get('event_type')`

lsh-0: it's a good point, I'm not testing against anything coming from elife-tools. Let me double check what elife-tools is returning

lsh-0: here is your work: https://github.com/elifesciences/issues/issues/6696#issuecomment-850580729-permalink

and it does indeed look like `event_type`, good catch

gnott:
I'm not sure how the output fits into Lax or the API schema, otherwise looking good!

src/main.py
lsh-0: test updated using actual output

lsh-0:
> I'm not sure how the output fits into Lax or the API schema, otherwise looking good!

Regarding the api-raml, articles have a 'article-history' API endpoint that previously was only POA and VOR snippets, but now contain a preprint stub as well. This should only affect applications requesting the latest (or an unspecified) version of the article-history endpoint.

I could have added the preprint information to the header of what the article-history endpoint returns rather than a new item in the list, but I wasn't getting any concrete answers about it's spec. 

With Lax I just need to figure out how to store this preprint information, serve it up and version the endpoint

lsh-0:
test-lax finally made it through testing with that other bot-lax change, merging this in now

PR #750, opened 2021-06-25
upgrades pip, installs wheel, falling back to legacy resolver

lsh-0:
bot-lax is currently depending on it's virtualenvs that are using a pretty old version of pip.

this upgrades pip, installs wheel so we get binary downloads and tells pip to use the old resolver behaviour, just until we port the dependencies to Pipenv like the other projects.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ et3==1.5.1
 git+https://github.com/elifesciences/elife-tools.git@31509bfc9a8b84656ed6a1e8433894c56eff0b83#egg=elifetools
 Flask==1.0.2
 Flask-Testing==0.7.1
-flex==6.13.2
+flex==6.14.1
 future==0.16.0
 isbnlib==3.9.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -6,7 +6,7 @@ et3==1.5.1
 git+https://github.com/elifesciences/elife-tools.git@31509bfc9a8b84656ed6a1e8433894c56eff0b83#egg=elifetools
 Flask==1.0.2
 Flask-Testing==0.7.1
-flex==6.13.2
```
lsh-0, line 9: I was getting this:

```
flex 6.13.2 requires click<7,>=3.3, but you'll have click 7.0 which is incompatible.
flex 6.13.2 requires PyYAML<4,>=3.11, but you'll have pyyaml 5.4 which is incompatible.
```

lsh-0:
feel free to merge this @gnott if the changes to the venv carry over into any other testing.

gnott (approved):
Looks good, no rush to merge for my purposes, thanks.

PR #763, opened 2021-07-20
Change preprint attribute type to event_type.

gnott:
Re issue https://github.com/elifesciences/issues/issues/6804

Change `type` to `event_type` where the preprint data is filtered from `pub_history()` when using the `elifetools` XML parser.


lsh-0:
cheers @gnott , looks fine and I've merged it

PR #773, opened 2021-08-27
glencoe, msid for new kitchen sink is preserved

lsh-0:
related changes in elife-spectrum: https://github.com/elifesciences/elife-spectrum/pull/302

### DIFF
--- a/src/glencoe.py
+++ b/src/glencoe.py
@@ -103,7 +103,7 @@ def expand_videos(msid, video):
     gc_id_str = ", ".join(gc_data.keys())
 
     v_id = video['id']
-    ensure(gc_data, "glencoe doesn't know %r, it doesn't have any media" % (v_id,))
+    ensure(gc_data, "glencoe doesn't know %r, it doesn't have any media for: %s" % (v_id, msid,))
     ensure(v_id in gc_data, "glencoe doesn't know %r, only %r" % (v_id, gc_id_str))
 
     video_data = gc_data[v_id]
--- a/src/main.py
+++ b/src/main.py
@@ -302,7 +302,11 @@ def expand_videos(data):
     def pred(element):
         return isinstance(element, dict) and element.get("type") == "video"
 
-    return visit(data, pred, partial(glencoe.expand_videos, utils.video_msid(msid)))
+    def fn(element):
+        new_msid = utils.video_msid_2(msid, element.get('uri'))
+        return glencoe.expand_videos(new_msid, element)
+
+    return visit(data, pred, fn)
 
 def expand_placeholder(msid, data):
 
--- a/src/tests/test_utils.py
+++ b/src/tests/test_utils.py
@@ -36,11 +36,42 @@ class Utils(BaseCase):
         self.assertEqual(stdout, 'pants-party')
 
     def test_video_msid(self):
-        self.assertEqual(9560, utils.video_msid(9560))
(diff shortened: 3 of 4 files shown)


lsh-0:
yes, elife-bot seems to have copy+pasted logic from bot-lax. I've opened a PR for elife-bot here: https://github.com/elifesciences/elife-bot/pull/1279

> This also matches the error on an end2end test run for PR elifesciences/elife-spectrum#302, the msid of the article which failed was 5453442605591467890 and is why it was checking for video metadata for an article with msid of 67890, the last 5 digits of the full msid.

There is actually a more subtle bug in here. The kitchen sink ID is being truncated by elife-spectrum when an ID is generated. I guess I was expecting:

    5453442601234567890

instead of 

    5453442605591467890

Not sure if it would affect the tests though if it's only videos we're looking for, and the videos have the kitchen sink msid preserved instead of replaced with the placeholder.

lsh-0:
> The kitchen sink ID is being truncated by elife-spectrum when an ID is generated.

fixed.

gnott:
Will this also allow a generated msid of `91234567890` to check for Glencoe metadata, or is a randomly generated kitchen sink msid not that target of the additional logic here?

lsh-0:
'91234567890' would be a randomly generated article ID by elife-spectrum (it has a leading '9'). 

Currently the kitchen sink XML will *not* have it's video IDs replaced with jinja placeholders:  https://github.com/elifesciences/elife-spectrum/pull/302/files#diff-6e8205ef8b50bde0ca86444c752db0385ae881ded5f567f7c439459a11209604R13-R20

But now that the truncation logic is removed for the kitchen sink from elife-bot and bot-lax (this PR) this shouldn't be a problem.

edit: I see you [changed my logic in elife-bot](https://github.com/elifesciences/elife-bot/pull/1279/files#diff-0ad5dd9b396711cf3968226f23fd3aa3e31df0f2acfedc1163b119251b92c430R93-R95) to re-introduce the truncation. That shouldn't matter, it just won't ever be truncated - unless you're using that `glencoe_checl.check_msid` function for anything other than glencoe.

gnott:
The msid truncation remains for `elife-bot` to check and download Glencoe resources because it currently uses the article's msid to look up metadata, where the actual video XML values are not as important. It's only used there for videos, as it was previously, and the msid remains unchanged in other places. How Lax ingests the XML should be unrelated I think.

lsh-0:
> How Lax ingests the XML should be unrelated I think.

Lax relies on upstream (bot-lax et al) to serve up the article-json ready to go.

gnott:
To clarify, I meant more like how the elife-bot Glencoe check and how bot-lax-adaptor works are very much unrelated, except elife-bot has copied JPG files to the bucket folder for each video, using the original msid as tested, and it doesn't use any truncated msid value in that situation.

I assumed the code introduced here was to make it easier to generate JSON and validate it when looking up Glencoe metadata directly perhaps?

lsh-0:
the code here it to remove the kitchen sink from having it's msids from being truncated after it changed from 666 to 1234567890. They don't need to be truncated as the kitchen sink template in elife-spectrum no longer has it's video elements replaced with jinja placeholders and has it's own set of videos on Glencoe.

The *other* templates in elife-spectrum (kitchen sink is just one) must have the original truncation behaviour preserved as those template articles still rely on video data from the *original* article.

PR #897, opened 2022-11-28
reviewed-preprints, events are now captured as regular preprint events.

lsh-0:


gnott (commented):
In following how it's invoked in `main.py`, I'm wondering about the statement,

```python
('preprint', [jats('pub_history'), preprint_events, first, to_preprint, discard_if_none_or_empty])
```

where it includes `first` in the function list. Will this return only one or multiple values? The test case for `preprint_events()` may not consider the usage here?

gnott:
I think there will be only one of the type `preprint`, so maybe it can be filtered further downstream?

PR #899, opened 2022-11-29
reviewed-preprint-list

lsh-0:
adds 'reviewed-preprint-list' to the '-history' map


gnott (approved):
I find no fault here, assuming the data will be handled by `lax` to produce the final list of preprint data exposed by the API gateway. 👍

PR #924, opened 2023-02-14
main.py, fixes parsing of doi in related article results

lsh-0:


gnott (approved):
I like the test case examples and looks good to me

PR #967, opened 2023-08-14
support related reviewed preprints

lsh-0:
- [x] get build back down to ~21mins
- [x] review

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -26,7 +26,7 @@ DYNCONFIG = configparser.ConfigParser(**{
     # these can be used like template variables
     # https://docs.python.org/2/library/configparser.html
     'defaults': {'dir': PROJECT_DIR}})
-DYNCONFIG.read(join(PROJECT_DIR, CFG_NAME)) # "/path/to/lax/app.cfg"
+DYNCONFIG.read(join(PROJECT_DIR, CFG_NAME)) # "/path/to/bot-lax-adaptor/app.cfg"
 
 def cfg(path, default=0xDEADBEEF):
     lu = {'True': True, 'true': True, 'False': False, 'false': False} # cast any obvious booleans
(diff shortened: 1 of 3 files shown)


src/conf.py
```diff
@@ -87,10 +92,15 @@ def load(path):
 
 POA_SCHEMA = load('api-raml/dist/model/article-poa.v3.json')
 VOR_SCHEMA = load('api-raml/dist/model/article-vor.v7.json')
+RELATED_SCHEMA = load('api-raml/dist/model/article-related.v2.json')
 
 REQUEST_SCHEMA = load('request-schema.json')
 RESPONSE_SCHEMA = load('response-schema.json')
 
+API_HOST = cfg("api.host", "elifesciences.org")
+API_URL = f"https://{ENV}--gateway.{API_HOST}" # https://continuumtest--gateway.elifesciences.org
```
gnott, line 101: Formatted string usage is interesting

src/main.py
```diff
@@ -109,20 +115,68 @@ def display_channel_to_article_type(display_channel_list):
...
+    # brute force approach. check API for every related MSID.
+    # return list(filter(None, map(fetch, reference_msid_list)))
+
+    if not msid_list:
+        return []
+
+    # `elifetools` is modifying the soup as we access it.
+    # multiple accesses to `references_json` results in strange behaviour,
+    # it's also hugely *slow*.
+    # * 24271 v1, ref 'bib38': the 'date' disappears.
```
gnott, line 151: Sorry about this, I can see how this is not doing a good job. The `bib38` is converted to type `unknown`, because it would be lacking data as a `book`, however if the now converted `unknown` reference goes through the same check, it will probably be mangled.

src/main.py
```diff
@@ -109,20 +115,68 @@ def display_channel_to_article_type(display_channel_list):
...
+    # copy the soup before accessing `references_json`
+    references = parseJATS.references_json(copy.copy(soup))
+
+    # check each reference for any relation's msid prefixed with an 'RP'
+    reference_msid_list = []
+    for ref in references:
+        val = ref.get('pages')
+        # `val` may be a string or a dict. just check strings for now.
+        if val and isinstance(val, str) and val[:2] == "RP":
```
gnott, line 160: Probably a rare situation, a non-eLife article may have an elocation-id starting with `RP`

gnott: Although if is also must have the right `msid` it'll probably be fine

gnott (approved):
A good read, I have no other comments thanks!

